PR #7, opened 2017-08-09
Update Nomad dependencies

### LINKED ISSUE #4: leader property not supported on tasks
The Nomad provider does not support the `leader` property on tasks. If the same `jobspec` is submitted to nomad directly it works.

### Terraform Version
`Terraform v0.9.11`

### Affected Resource(s)
- nomad_job

```hcl
provider "nomad" {
  address = "127.0.0.1: 4646"
}

resource "nomad_job" "test" {
  jobspec = <<EOF
job "test-job" {
  type = "service"
  datacenters = ["dc1"]

  group "test-group" {
    count = "1"

    task "test-task" {
      driver = "docker"

      config {
        image = "hello-world"
      }

      leader = true
    }
  }
}
EOF
}
```

### Debug Output
https://gist.github.com/cstruct/3630a5c93044d579d66898c0a07e0eb5

### Expected Behavior
That Terraform schedules a task that is the leader in the group.

### Actual Behavior
Terraform fails and says leader is an invalid key.

### Steps to Reproduce
1. `terraform apply`

blalor:
This updates the vendored Nomad dependencies to v0.6.0, with a couple of other transitive dependencies as specified in Nomad's vendor.json.  

I struggled mightily with govendor and resorted to brute-forcing the fetch of the previously-existing vendored sources under github.com/hashicorp/nomad (plus the others).  This does not feel like the correct approach. 🤔 However, the end result works.  I believe this project's vendor.json needs to be adjusted to ease future maintenance.

Fixes #4.


dadgar:
@grubernaut Gave it a look and from Nomad perspective looks 👍

cstruct:
Ping @radeksimko. What's necessary for this to land? It would make my day if this landed soon.

nomad/resource_job.go
```diff
@@ -54,16 +51,13 @@ func resourceJobRegister(d *schema.ResourceData, meta interface{}) error {
...
 	// Parse it
-	jobspecStruct, err := jobspec.Parse(strings.NewReader(jobspecRaw))
+	job, err := jobspec.Parse(strings.NewReader(jobspecRaw))
 	if err != nil {
 		return fmt.Errorf("error parsing jobspec: %s", err)
 	}
 
-	// Initialize and validate
-	jobspecStruct.Canonicalize()
-	if err := jobspecStruct.Validate(); err != nil {
```
radeksimko: I understand the validation is now part of `Parse`, right? So it's just implicit, rather than explicit? 🤔

sirocode:
Why you consider this issue as "enhancement"? It's a bug.

cstruct:
@sirocode I think this was marked as an enhancement when the only reported issue about this was missing support for the `leader` property. It could be argued adding support for a new property where everything else works is an enhancement and not a bug.

sirocode:
@cstruct I understand, I just can not wait to finally see this issue fixed!

nomad/resource_job.go
dadgar: @radeksimko It is not validated as part of `Parse()`. The job is validated during Register. If it is invalid and error will be returned. If you want early validation,  you would have to call validate before register: https://github.com/hashicorp/nomad/blob/master/api/jobs.go#L42

blalor:
I'm on vacation right now, but I'll try to address the issues raised on Monday. Sorry for the delay!

cstruct:
@sirocode If you can't wait you can build it from source and bundle it with `terraform-bundle`, that's what I've done.

apparentlymart:
hashicorp/nomad#1982 was opened to track the issue that we weren't able to use the upstream Nomad due to the need for all of the drivers to be present.

Since that issue remains open, I assume that limitation still applies and so we need to retain the local patches in Terraform to work around the problem. :disappointed:

dadgar:
@apparentlymart Sorry about that, I just closed the issue. That has been resolved since Nomad v0.5.5

radeksimko:
@dadgar Thanks for the clarification.

@blalor My only question remains - can we remove `github.com/hashicorp/nomad/` from `ignore` then?

I'm happy to do it for you, if you don't have time to spend on this anymore - just let me know.

nomad/resource_job.go
radeksimko: Understood. So the user practically won't see any difference, because it's validated in CRUD - in either case during `apply`. If we wanted to do early validation we'd do it in `ValidateFunc` (which runs during `plan`).

Still a little bit confusing to see this being removed as part of an upgrade, but it doesn't do any harm, I guess.

justinpersaud:
any rough eta on when this might be merged? hoping to leverage the latest nomad features through terraform

chandy:
hi @blalor is there any chance you could make the update @radeksimko is proposing or just give him the goahead to do it?  We would like very much to use this feature.  Thanks for your work on it so far!

nomad/resource_job.go
blalor: It was removed because `jobspec.Parse` returns `*api.Job` which does not have a `Validate` function.  That was the simple solution.  Based on the conversation I don't believe there's much value in performing additional validation, but I'm happy to if that's what you'd like.

radeksimko:
Thanks for making those changes. 👍

chandy:
Thank you @blalor !

PR #8, opened 2017-08-10
vendor: github.com/hashicorp/terraform/...@v0.10.0

radeksimko:
```
govendor fetch github.com/hashicorp/terraform/...@v0.10.0
```


radeksimko:
@catsby Probably not - even though the tests are green I removed that commit, just to be safe and avoid rolling out multiple risky changes at the same time.

PR #13, opened 2017-10-05
Add a vault_token attribute on the provider

### LINKED ISSUE #2: Allow specification of VAULT_TOKEN for nomad_job resource
_This issue was originally opened by @dansteen as hashicorp/terraform#14664. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

### Terraform Version
Terraform v0.9.2

### Affected Resource(s)
nomad_job

Nomad is able to be configured to require that anyone creating a job pass in a token proving that they have access to the vault data that that job requires (see here: https://www.nomadproject.io/docs/agent/configuration/vault.html#allow_unauthenticated )

It would be great if the terraform nomad_job resource allowed us to specify the token to use for that authentication.   

Thanks!

lethalpaga:
In order to run jobs that specify vault policies, a vault token must be provided.
This adds a vault_token field in the provider that gets injected in the job file before sending it to nomad.

I'm not sure how to write better acceptance tests. At the moment the test will fail unless a valid vault token is added to either the provider, $VAULT_TOKEN or ~/.vault-token

Should fix #2


ryanmickler:
This is great, i've been needing this for a while!

lethalpaga:
I fixed the test issue by adding a `testintegration` target in the Makefile. It starts a nomad agent and a vault server in `-dev` mode with a known token before running the tests.

lethalpaga:
@radeksimko anything I can do to help get this merged?

apparentlymart:
Hi @lethalpaga! Sorry we've let this sit here so long.

Our attention (that is, of the Terraform team at HashiCorp) has been elsewhere for the last few months and so we've not been paying close attention to the Nomad provider. However, we do want to get this feature in so we'll come back and take a look at this soon, along with some other investment here.

GNUmakefile
```diff
@@ -14,6 +14,11 @@ test: fmtcheck
 testacc: fmtcheck
 	TF_ACC=1 go test $(TEST) -v $(TESTARGS) -timeout 120m
 
+testintegration: fmtcheck
```
paddycarver: Any reason to add this to the makefile? Is `testacc` not sufficient?

nomad/provider.go
```diff
@@ -52,6 +68,36 @@ func Provider() terraform.ResourceProvider {
 	}
 }
 
+// populateTokenPath figures out the token path using homedir to get the user's
+// home directory
```
paddycarver: Is this a file/pattern taken from elsewhere? From what I can tell, Vault doesn't do this in the CLI. I'd like to try to match as closely as possible what Vault does, I think, if for no other reason than it doesn't encourage developers to store sensitive data on their hard drives/in their CI systems.

nomad/provider.go
```diff
@@ -60,10 +106,23 @@ func providerConfigure(d *schema.ResourceData) (interface{}, error) {
 	config.TLSConfig.ClientCert = d.Get("cert_file").(string)
 	config.TLSConfig.ClientKey = d.Get("key_file").(string)
 
+	// Get the vault token from the config, VAULT_ADDR
```
paddycarver: nit: `VAULT_TOKEN`

nomad/resource_job_test.go
```diff
@@ -329,7 +409,11 @@ job "bar" {
             logs {
                 max_files = 3
                 max_file_size = 10
-            }
+			}
+			
+			vault {
+				policies = ["default"]
+			}
```
paddycarver: looks like we have some tabs/spaces stuff going on here

nomad/resource_job_test.go
```diff
@@ -305,22 +367,40 @@ func testResourceJob_updateCheck(s *terraform.State) error {
...
+				Check:  testResourceJob_initialCheck,
+			},
+		},
+		CheckDestroy: testResourceJob_checkDestroy("test"),
+	})
+}
+
+var testResourceJob_vaultConfig = `
+provider "nomad" {
+	vault_token = "terraform-provider-nomad-token"
```
paddycarver: This (obviously) isn't a valid token. Is the intent here to get a working token passed in? Is there any reason not to use the environment variable configuration option?

nomad/resource_job_test.go
```diff
@@ -305,22 +367,40 @@ func testResourceJob_updateCheck(s *terraform.State) error {
...
 
-var testResourceJob_parameterizedJob = `
+func TestResourceJob_vault(t *testing.T) {
+	r.Test(t, r.TestCase{
+		Providers: testProviders,
+		PreCheck:  func() { testAccPreCheck(t) },
+		Steps: []r.TestStep{
+			{
+				Config: testResourceJob_vaultConfig,
+				Check:  testResourceJob_initialCheck,
```
paddycarver: Is there any way we could check the output for Vault-specific things, to ensure the token was actually used?

lethalpaga:
Alright so I think I've addressed @paddycarver's concerns, but I'm struggling a bit vendoring the vault dependencies for the token lookup. It pulls in heaps of unnecessary stuff and fails to compile (https://travis-ci.org/terraform-providers/terraform-provider-nomad/builds/337878811). I don't have much time to work on that atm, so help would be much appreciated.
Ideally the hashicorp/vault/command package should be broken down in smaller packages as this would only require utils.go and config.go. . Otherwise I can revert to the old way for the time being and get this merged.

Other than that: 

* The provider now uses vault's builtin token helpers. The cool thing with that is that it now also works if you have custom helpers. The bad thing is that it pulls in a lot of dependencies.
  The original code I had came from the vault terraform provider, it would be worth updating that one as well if it hasn't been done already.

* I renamed the testintegration target to localtestacc to make it more meaningful. The initial intent was to make running those tests easier. Unfortunately it's becoming a bit complex since the addition of the ACL resources. I spent some time trying to configure that script but didn't manage to make it work (I haven't played with nomad ACLs before so I'd need to do some more reading about this).

nomad/provider.go
```diff
@@ -69,10 +94,23 @@ func providerConfigure(d *schema.ResourceData) (interface{}, error) {
 	config.TLSConfig.ClientKey = d.Get("key_file").(string)
 	config.SecretID = d.Get("secret_id").(string)
 
+	// Get the vault token from the config, VAULT_TOKEN
+	// or ~/.vault-token (in that order)
+	var err error
+	vaultToken := d.Get("vault_token").(string)
+	if vaultToken == "" {
+		vaultToken, _ = getToken()
```
paddycarver: Why is this error being ignored? If it's because we don't want to error out on a file that's optional, I get that. From the default implementation[1], it looks like it doesn't return an error if the file doesn't exist, so this should work as expected if no file is detected, but we could/should still return any error we run across, as it generally would indicate the user had settings, but we failed to retrieve/parse them. Which, as a user, I'd like to know about, rather than silently failing.

nomad/resource_job.go
```diff
@@ -56,6 +56,11 @@ func resourceJobRegister(d *schema.ResourceData, meta interface{}) error {
 		return fmt.Errorf("error parsing jobspec: %s", err)
 	}
 
+	// Inject the Vault token if provided
+	if providerConfig.vaultToken != nil {
```
paddycarver: This check isn't really necessary, I don't think; setting the VaultToken to nil would have the same effect as not setting it at all.

lethalpaga: I was sure I had issues with that but it seems to be working without the condition now

lethalpaga:
Good points, I fixed the error handling, it was all quirks from the old code.
If you don't mind taking care of the vendoring part that would be awesome

paddycarver:
Awesome, this looks great! Thanks so much for taking the initiative on this while I've been super slow. I'll get the dependencies on this in order, then get it merged in.

paddycarver:
Hi @lethalpaga,

As promised, an update. I worked with the Vault team to get hashicorp/vault#4055 merged, which separates out the portions of the `github.com/hashicorp/vault/command` package we're actually using. This makes vendoring a bit more reasonable, because `command` imported basically all of Vault, and vendoring that was tricky.

I'm going to open and merge a new PR that's based off your commits before you vendored things, then cherry-picking non-vendor commits, and finally adding my own vendoring commit, which will resolve the situation, I believe. It should retain your authorship of all your commits.

Sorry for the runaround and complexity on this! We got to a good solution though, I think.

Thanks for all the hard work and patience.

paddycarver:
This has been merged now. Thanks so much!

PR #17, opened 2017-12-01
Add support for ACL secret_id

edio:
acctest was failing even before the changes, first commit fixes that (in the easiest way I could figure out)

nomad api had to updated to 0.7.0 for this.
It's basically `govendor fetch ...@=v0.7.0`

### DIFF
--- a/nomad/provider_test.go
+++ b/nomad/provider_test.go
@@ -4,6 +4,8 @@ import (
 	"os"
 	"testing"
 
+	"github.com/hashicorp/nomad/api"
+	"github.com/hashicorp/terraform/helper/resource"
 	"github.com/hashicorp/terraform/helper/schema"
 	"github.com/hashicorp/terraform/terraform"
 )
@@ -14,7 +16,7 @@ import (
 //   and extract the "nomad" binary
 //
 // - Run the following to start the Nomad agent in development mode:
-//       nomad agent -dev
+//       nomad agent -dev -acl-enabled
 //
 // - Run the Terraform acceptance tests as usual:
 //       make testacc TEST=./builtin/providers/nomad
@@ -29,11 +31,16 @@ func TestProvider(t *testing.T) {
 
 var testProvider *schema.Provider
 var testProviders map[string]terraform.ResourceProvider
+var managementToken string
 
 func init() {
 	testProvider = Provider().(*schema.Provider)
 	testProviders = map[string]terraform.ResourceProvider{
 		"nomad": testProvider,
+	}
+
+	if v := os.Getenv(resource.TestEnvVar); v == "1" {
+		setupAcl()
 	}
 }
 
@@ -41,4 +48,26 @@ func testAccPreCheck(t *testing.T) {
 	if v := os.Getenv("NOMAD_ADDR"); v == "" {
 		os.Setenv("NOMAD_ADDR", "http://127.0.0.1:4646")
 	}
+	os.Setenv("NOMAD_TOKEN", managementToken)
+}
+
+func testAccPreCheckAnonymous(t *testing.T) {
+	if v := os.Getenv("NOMAD_ADDR"); v == "" {
+		os.Setenv("NOMAD_ADDR", "http://127.0.0.1:4646")
+	}
+	os.Unsetenv("NOMAD_TOKEN")
+}
+
+func setupAcl() {
+	client, err := api.NewClient(&api.Config{})
+	if err != nil {
+		panic(err)
+	}
+
+	acl, _, err := client.ACLTokens().Bootstrap(nil)
+	if err != nil {
+		panic(err)
+	}
+
+	managementToken = acl.SecretID
 }
--- a/nomad/resource_job_test.go
+++ b/nomad/resource_job_test.go
@@ -243,8 +243,7 @@ job "bar" {
         task "foo" {
             driver = "raw_exec"
             config {
-                command = "/bin/sleep"
-                args = ["1"]
+                command = "/bin/true"
             }
 
             resources {
@@ -308,7 +307,7 @@ func testResourceJob_updateCheck(s *terraform.State) error {
 var testResourceJob_parameterizedJob = `
 resource "nomad_job" "test" {
     jobspec = <<EOT
-job "bar" {
+job "parameterized" {
     datacenters = ["dc1"]
     type = "batch"
     parameterized {
--- a/nomad/resource_job_test.go
+++ b/nomad/resource_job_test.go
@@ -10,7 +10,21 @@ import (
 	"github.com/hashicorp/terraform/terraform"
 
 	"github.com/hashicorp/nomad/api"
+	"regexp"
 )
+
+func TestResourceJob_basic_anonymous(t *testing.T) {
+	r.Test(t, r.TestCase{
+		Providers: testProviders,
+		PreCheck:  func() { testAccPreCheckAnonymous(t) },
+		Steps: []r.TestStep{
+			{
+				Config:      testResourceJob_initialConfig,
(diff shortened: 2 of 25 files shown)


apparentlymart:
Thanks for working on this, @edio!

We can't look at this immediately due to other work going on but we should be working on this provider more in the near future so we'll plan to review/test this as part of other work. Sorry for the delay...

Ashald:
@apparentlymart we would highly appreciate if you could reconsider that and take a look at this pull request. At this moment Nomad provider for Terraform is *literally unusable* with upstream version of Nomad once you enable ACL (which is an obvious thing to do in any production-like environment). Also, given that the provider does validate the job definitions we essentially cannot use any of the job features added since version `0.6`. Regardless of huge improvements that are planned for the provider it's a disaster to not be abel to use official version of the provider with the official version of Nomad.

GNUmakefile
```diff
@@ -12,6 +12,8 @@ test: fmtcheck
 		xargs -t -n4 go test $(TESTARGS) -timeout=30s -parallel=4
 
 testacc: fmtcheck
+	@echo "Make sure nomad is running"
+	@echo "	nomad agent -dev -acl-enabled"
```
paddycarver: As a rule of thumb, we tend to have the environment setup required for our acceptance tests happen outside the `testacc` target. This is so our CI and contributors have the option to reuse stable environments, and to avoid hardcoding the environment in.

CHANGELOG.md
```diff
@@ -1,4 +1,10 @@
-## 1.0.1 (Unreleased)
+## 1.1.0 (Unreleased)
+
+IMPROVEMENTS:
+
+* provider: Update Nomad to v0.7.0
+* provider: Allow setting ACL in hcl
+
```
paddycarver: In this case it doesn't really matter, but we tend to update the CHANGELOG as a separate step, outside the PR process. Otherwise, merge conflicts quickly get out of control.

nomad/provider_test.go
```diff
@@ -29,16 +30,38 @@ func TestProvider(t *testing.T) {
...
 var testProviders map[string]terraform.ResourceProvider
+var managementToken string
 
 func init() {
 	testProvider = Provider().(*schema.Provider)
 	testProviders = map[string]terraform.ResourceProvider{
 		"nomad": testProvider,
 	}
+
+	if v := os.Getenv("TF_ACC"); v == "1" {
```
paddycarver: nit: [`resource.TestEnvVar`](https://godoc.org/github.com/hashicorp/terraform/helper/resource#TestEnvVar) can be used to avoid hardcoding this.

nomad/provider_test.go
```diff
@@ -29,16 +30,38 @@ func TestProvider(t *testing.T) {
...
+
+func testAccPreCheckAnonymous(t *testing.T) {
+	if v := os.Getenv("NOMAD_ADDR"); v == "" {
+		os.Setenv("NOMAD_ADDR", "http://127.0.0.1:4646")
+	}
+	os.Unsetenv("NOMAD_TOKEN")
+}
+
+func setupAcl() {
+	client, _ := api.NewClient(&api.Config{})
```
paddycarver: It looks like we're ignoring an error here. I'd be a lot more comfortable if we checked and panicked on it.

nomad/resource_job_test.go
```diff
@@ -208,6 +223,7 @@ func testResourceJob_checkExists(s *terraform.State) error {
 
 func testResourceJob_checkDestroy(jobID string) r.TestCheckFunc {
 	return func(*terraform.State) error {
+		time.Sleep(time.Second) // let things happen in background
```
paddycarver: What things? What are we waiting for here? Why is this necessary?

Sometimes Sleeps are necessary, and it's always unfortunate when that happens, but if we document why, at least we can remove them in the future when they stop being necessary.

edio: This test is failing on the recent commit of the master branch of this repo for me.

I'm not sure, whether this test was written in such way itnentionally. I might have misunderstood the whole idea of this test.

But, anyway, my understanding is:
* in test we create a job with command `/bin/sleep 1`
* right after we create the job we check that the job is "dead"
* apparently, by the time we check job state, it is still not "dead", so test fails

So I introduced sleep before we check.

Perhaps, I'd better change job definition to `/bin/true`, so the test will pass (in most cases, explained below) without introducing delay before the check.

This solution will practically work, but is still conceptually wrong and hypothetically may fail on a very slow machine.

But, I don't see, how the previous version of the test have been working at all.

Let me know, what do you think.

GNUmakefile
edio: If I understood you correctly, you're suggesting, that `testacc` should depend on, say, `run_nomad`, which will start it for the user, correct?

If that's the case, I totally agree with you, but I just kept original approach.

Please confirm if I udnerstood you correctly, and I'll fix.

edio:
I started responding from bottom to top, and finally got to your main question :)

My knowledge is probably way more limited than yours :) But here's what I think.

With these acceptance tests we have 2 mutually exclusive options:
1. user runs nomad, bootstraps ACL, extracts token manually and passes it to tests with env var
2. we do all, or part of this, automactically (this is also relevant to your suggestion in the Makefile)

Why these are exclusive: when we bootstrap ACL we can't just tell nomad to use our own token for that, token is being generated by nomad automatically. Hence, we either do it manually, or we do it in init.

I didn't really understand your concern about parallel tests, sorry.

I'll just emphasize one thing, maybe it'll clarify something: bootstrap ACL token and nomad instance is a single invariant. If we bootstrapped nomad instance, token won't change. So if your CI is going to run couple of different tests against the same instance of nomad, it, of course, will fail.

I saw the original comment on this test, that test expects fresh instance of nomad, so I concluded, that this is something I can safely rely on.

nomad/resource_job_test.go
paddycarver: Appreciate the clarification, and thanks for fixing the test!

I'm uncomfortable with the Sleep, and think a better idea would probably be to introduce some polling to make things a bit more robust (and agree that `/bin/true` is probably a better test job to run), but it's a bit removed from the main problem this PR is addressing, so I'm ok just letting it pass for the moment and circling back around to fix it after we get this resolved.

GNUmakefile
paddycarver: Basically what I'm suggesting is that `testacc` not try to start Nomad at all. Instead, the user running `testacc` should start Nomad themselves, _then_ run `testacc`. A Makefile target can be added to run Nomad, but I think "running the tests" and "setting up the environment the tests run against" should be two separate commands.

edio: Oh I see, so it's the opposite of what I understood :)

Well, this does not setup the environment, but just merely a reminder for the user. But now that you pointed this out, I think, that a right place for this is README.md.

I'm going to address all this today.

edio:
@paddycarver I addressed all your comments in the code.
Do you still want me to change the code, so the token for tests is passed as env var?
The downside of that would be more work for user or CI to setup environment.

paddycarver:
@edio I think this would be better with the token passed as an env var, and that seems more inline with how our providers tend to operate. I think we could _also_ have a script that automates the setup and token retrieval, but I think we'd be best served by keeping that separate from the running of the tests. I'm not quite sure _why_ the comment says the cluster needs to be empty, but I think if we can avoid reenforcing that requirement where possible, that's probably for the best.

Our support engineers, @angrycub in particular, mentioned that you'd prefer I just go ahead and make the changes and merge it in over making them yourself, and I'm willing to do that. I'll go ahead and do that now, but I do appreciate your work, effort, and patience on this!

PR #22, opened 2018-01-29
Add a resource for managing ACL policies.

paddycarver:
Add a new resource that enables the management of ACL policies. It
allows declaratively defining an ACL policy, and Terraform will make
sure it exists in the Nomad cluster.

### DIFF
--- a/nomad/resource_acl_policy.go
+++ b/nomad/resource_acl_policy.go
@@ -1,0 +1,140 @@
+package nomad
+
+import (
+	"fmt"
+	"log"
+	"strings"
+
+	"github.com/hashicorp/nomad/api"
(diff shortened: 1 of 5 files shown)


nomad/resource_acl_policy.go
```diff
@@ -0,0 +1,140 @@
...
+	log.Printf("[DEBUG] Deleted ACL policy %q", name)
+
+	return nil
+}
+
+func resourceACLPolicyRead(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*api.Client)
+	name := d.Id()
+
+	// retrie the policy
```
mbfrahry: retrie to retrieve?

PR #23, opened 2018-01-29
Vendor the acctest package.

paddycarver:
We're using acctest for our tests now, so let's get it vendored.

Fixes Travis build.


bflad:
We'll need to pull in related dependencies as well, thanks `govendor`! 😦  
`vendor/github.com/hashicorp/terraform/helper/acctest/random.go:16:2: cannot find package "golang.org/x/crypto/ssh"`

paddycarver:
This is why we can't have nice things.

PR #24, opened 2018-01-31
Add regions data source for Nomad.

paddycarver:
This gives us a list of the regions available in Nomad, which allows for
things like deploying a job in every region in the cluster.


paddycarver:
The CI failure is fixed in #23.

nomad/data_source_regions.go
```diff
@@ -0,0 +1,37 @@
...
+			"regions": {
+				Type:     schema.TypeList,
+				Elem:     &schema.Schema{Type: schema.TypeString},
+				Computed: true,
+			},
+		},
+	}
+}
+
+func genericSecretDataSourceRead(d *schema.ResourceData, meta interface{}) error {
```
paultyng: this name seems wrong?

paddycarver: Oops. Fixed.

PR #25, opened 2018-02-12
Manage quota specifications.

paddycarver:
Add a resource to manage quota specifications, limiting the resources
that can be used in a region.

The tests for this are 🙀 because sets nested in sets, but... they gave
me confidence that it works? 🤷


nomad/resource_quota_specification.go
```diff
@@ -0,0 +1,270 @@
...
+
+	spec := api.QuotaSpec{
+		Name:        d.Get("name").(string),
+		Description: d.Get("description").(string),
+	}
+	limits, err := expandQuotaLimits(d)
+	if err != nil {
+		return err
+	}
+	spec.Limits = limits
```
paultyng: Not a big deal, but this could be moved up to that you just assign during the creation of the spec and not additionally here. Same as in `...Update`.

nomad/resource_quota_specification.go
```diff
@@ -0,0 +1,270 @@
...
+
+	// upsert our quota spec
```
paultyng: Since this is an upsert like the update, could you just called the update here and then `SetId`? Or add some shared inner func? They appear to be the same code except for some messaging and the `SetId`.

nomad/resource_quota_specification.go
```diff
@@ -0,0 +1,270 @@
...
+func resourceQuotaSpecificationExists(d *schema.ResourceData, meta interface{}) (bool, error) {
+	client := meta.(*api.Client)
+
+	name := d.Id()
+	log.Printf("[DEBUG] Checking if quota specification %q exists", name)
+	resp, _, err := client.Quotas().Info(name, nil)
+	if err != nil {
+		// As of Nomad 0.4.1, the API client returns an error for 404
+		// rather than a nil result, so we must check this way.
+		if strings.Contains(err.Error(), "404") {
```
paultyng: This is super ugly :cry:  is there an issue open on the Nomad API to at least return a named/typed err?

paddycarver: There is! https://github.com/hashicorp/nomad/issues/1849

PR #26, opened 2018-02-12
Implement namespace management.

paddycarver:
Allow Terraform to manage Nomad namespaces. At the moment, we have an
unadvertised parameter for setting a quota on the namespace, as well.
This will be tested and then advertised in a future commit, after quota
management with Terraform (#25) is merged.


nomad/resource_namespace.go
```diff
@@ -0,0 +1,144 @@
...
+		Name:        d.Get("name").(string),
+		Description: d.Get("description").(string),
+		Quota:       d.Get("quota").(string),
+	}
+
+	// upsert our namespace
+	log.Printf("[DEBUG] Creating namespace %q", namespace.Name)
+	_, err := client.Namespaces().Register(&namespace, nil)
+	if err != nil {
+		return fmt.Errorf("error inserting namespace %q: %s", namespace.Name, err.Error())
```
tombuildsstuff: _minor_ given the comment above, should this be `upserting`?

nomad/resource_namespace.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+	log.Printf("[DEBUG] Created namespace %q", namespace.Name)
+	d.SetId(namespace.Name)
+
+	return resourceNamespaceRead(d, meta)
+}
+
+func resourceNamespaceUpdate(d *schema.ResourceData, meta interface{}) error {
```
tombuildsstuff: _minor:_ given these methods are the same - we can combine them for now?

nomad/resource_namespace.go
```diff
@@ -0,0 +1,144 @@
...
+func resourceNamespaceCreate(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*api.Client)
+
+	namespace := api.Namespace{
+		Name:        d.Get("name").(string),
+		Description: d.Get("description").(string),
+		Quota:       d.Get("quota").(string),
+	}
+
+	// upsert our namespace
```
tombuildsstuff: _minor:_ I think we can remove this, since it's covered by the print statement below?

nomad/resource_namespace.go
```diff
@@ -0,0 +1,144 @@
...
+	log.Printf("[DEBUG] Updated namespace %q", namespace.Name)
+
+	return resourceNamespaceRead(d, meta)
+}
+
+func resourceNamespaceDelete(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*api.Client)
+	name := d.Id()
+
+	// delete the namespace
```
tombuildsstuff: _minor:_ I think we can remove this, since it's covered by the print statement below?

nomad/resource_namespace.go
```diff
@@ -0,0 +1,144 @@
...
+func resourceNamespaceExists(d *schema.ResourceData, meta interface{}) (bool, error) {
+	client := meta.(*api.Client)
+
+	name := d.Id()
+	log.Printf("[DEBUG] Checking if namespace %q exists", name)
+	resp, _, err := client.Namespaces().Info(name, nil)
+	if err != nil {
+		// As of Nomad 0.4.1, the API client returns an error for 404
+		// rather than a nil result, so we must check this way.
+		if strings.Contains(err.Error(), "404") {
```
tombuildsstuff: I'm assuming that means the status code also isn't a 404?

nomad/resource_namespace.go
```diff
@@ -0,0 +1,144 @@
...
+			return false, nil
+		}
+
+		return true, fmt.Errorf("error checking for namespace %q: %#v", name, err)
+	}
+	if resp == nil {
+		// just to be sure
+		return false, nil
```
tombuildsstuff: this probably wants a log statement here

website/docs/r/namespace.html.markdown
```diff
@@ -0,0 +1,32 @@
...
+  description = "Shared development environment."
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+- `name` `(string: <required>)` - A unique name for the namespace.
+- `description` `(string: "")` - A description of the namespace.
```
tombuildsstuff: I might be wrong, but isn't the convention:

```
* `name` - (Required) A unique name for the namespace. Changing this forces a new resource to be created.
* `description` - (Optional) Description for this namespace.
* `quota` - (Optional) Quota for this namespace.
```

(note: quota is missing above)

paddycarver: I think in general, yes, but when I got to the provider, it already had the above format, so I was trying to keep it consistent within the provider :/ Happy to do a sweep to update, though I have to admit I _do_ think the version I copied is a bit more informative, providing information on defaults and types.

nomad/resource_namespace.go
paddycarver: Good catch, thanks!

nomad/resource_namespace.go
paddycarver: It probably is, unfortunately the client library doesn't expose that to me :/

website/docs/r/namespace.html.markdown
paddycarver: Oh, also, quota is intentionally missing until #25 gets merged, because I have no way to test it until then. I just included the code here because I was already in the code, but I don't want to document it until I can actually test it.

nomad/resource_namespace.go
paultyng: Please link to https://github.com/hashicorp/nomad/issues/1849 in this comment so we can follow up later more easily.

PR #27, opened 2018-02-20
Implement managing sentinel policies.

paddycarver:
Add a `nomad_sentinel_policy` resource that manages a Sentinel policy
for the Nomad cluster.

Note this PR also runs `govendor add +external`, so, uh.... that's why
it's _massive_.


amanjeev:
Does this feature also include running sentinel policy tests on `terraform plan`?

paddycarver:
No, this feature is solely for registering the policies within Nomad. Unfortunately, because Sentinel is built into Nomad Enterprise, and not into Terraform, I don't see a good way to make that happen--unless Nomad Enterprise were to add a "would this job pass Sentinel if I submitted it?" endpoint which we could call as part of plan.

PR #28, opened 2018-02-20
 Add policy_override to job.

paddycarver:
Add a `policy_override` field to `nomad_job` to override any
soft-mandatory Sentinel policies that may fail during the job
registration. Depends on #27.

The only files changed in this PR that aren't from #27 being its
base are the changes in `nomad/resource_job.go`,
`nomad/resource_job_test.go`, and `website/docs/r/job.html.markdown`.
The rest can be safely ignored/reviewed in #27 instead.


paultyng:
Could you possibly do an initial `+external` run as its own commit to shrink that first commit a little? Or is it all the same dependencies needed in that one?

paddycarver:
All the dependencies are needed to make #27 work. #27 is needed to make this work.

This shouldn't have any new dependencies of its own.

PR #32, opened 2018-03-23
Adding datasource for nomad_job resource

slapula:
This is the first of a couple PRs I'd like to submit.  This one introduces a `nomad_job` datasource which allows a user to act or respond to the resource's various attributes.  I've tested this in my limited environments and it works well for me, but further testing and feedback would be greatly appreciated.


nomad/datasource_nomad_job.go
```diff
@@ -0,0 +1,287 @@
...
+
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func dataSourceJob() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceJobRead,
+		Schema: map[string]*schema.Schema{
+
+			"id": {
```
paddycarver: Naming things `id`, unfortunately, really messes with our provider framework. `job_id` may be a better choice here.

nomad/datasource_nomad_job.go
```diff
@@ -0,0 +1,287 @@
...
+			},
+			"task_groups": {
+				Description: "Job Task Groups",
+				Computed:    true,
+				Type:        schema.TypeList,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"placed_canaries": {
+							Type:     schema.TypeString,
+							Optional: true,
```
paddycarver: I think these should all be `Computed`, not `Optional`. I don't know what the behavior of an `Optional` field in a `Computed` block is, but the semantics of it don't really make sense. `Optional` indicates you can supply a value but aren't required to. `Computed` without `Optional` or `Required` prohibits supplying a value.

nomad/datasource_nomad_job.go
```diff
@@ -0,0 +1,287 @@
...
+						},
+						"timezone": {
+							Type:     schema.TypeString,
+							Optional: true,
+						},
+					},
+				},
+			},
+			"vault_token": {
+				Description: "Vault Token",
```
paddycarver: If this is doing what I think it's doing--returning the vault token that the job is using--it should be marked `Sensitive`.

nomad/datasource_nomad_job_test.go
```diff
@@ -0,0 +1,119 @@
...
+				Config: testAccCheckNomadJobConfig_basic(jobId),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckNomadJobExists("nomad_job.foobar", &testJob),
+				),
+			},
+			{
+				Config: testAccCheckNomadJobConfig_basic(jobId),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr(
+						"data.nomad_job.foobar", "id", fmt.Sprintf("%s", jobId)),
```
paddycarver: This could probably be combined with the first TestStep, and ideally would check more of the computed attributes to make sure they're set correctly.

nomad/datasource_nomad_job_test.go
```diff
@@ -0,0 +1,119 @@
...
+				rs.Primary.ID, err)
+		}
+	}
+
+	return nil
+}
+
+func testAccCheckNomadJobConfig_basic(str string) string {
+	return fmt.Sprintf(`
+data "nomad_job" "foobar" {
```
paddycarver: This test should fail, because it doesn't set up the job but expects it to exist.

website/docs/d/job.html.markdown
```diff
@@ -0,0 +1,57 @@
...
+* `job_modify_index`: Job modification index.
+* `stop`: Job enabled status.
+* `priority`: Job priority.
+* `parent_id`: Job's parent ID.
+* `task_groups`: Job's Task Groups.
+* `stable`: Job stability status.
+* `all_at_once`: All-at-once setting status.
+* `contraints`: Job constraints.
+* `update_strategy`: Job's update strategy.
+* `periodic_config`: Job's periodic configuration.
```
paddycarver: A bit more information (that these are blocks, what's in the blocks, etc.) would be useful here. Also, data types for all of these properties would be useful.

website/docs/d/job.html.markdown
```diff
@@ -0,0 +1,57 @@
...
+* `vault_token`: Job Vault token.
```
paddycarver: We should probably put a warning up front that using this resource introduces the Job's Vault token into state.

slapula: That's a good point.  I'm not quite sure of the value in including this data so I'm going to drop it.

slapula:
@paddycarver Thanks for the feedback!  I'm not completely satisfied with the documentation too so I've been working on that. I believe I've also fixed the tests so that they actually work and test reasonable values.

I'll admit that this datasource is something that I've only really used in my personal test environment.  The data I pull from this resource is used in a couple templated scripts that I've included in my Terraform code. On a broader level,  I'm a big fan of Terraform resources pulling in as much data as reasonably possible.  I could pull this from the cluster directly but I feel that if I'm managing the job via Terraform I should be pulling the data with Terraform as well.  I'm open to other arguments for or against this perspective however.

One of the things I'm thinking about right now is if this datasource needs to pull in every single bit of data from the API call. `nomad status <job_id>` doesn't display that much information and maybe the most appropriate direction here is to follow that standard. We could then make detailed task group data (for example) accessible through a different datasource.  Just an idea, not really sure how practical that would be.

PR #33, opened 2018-03-23
Adding datasource for nomad_deployment resource

slapula:
This one introduces a nomad_deployment datasource which allows a user to act or respond to this resource's various attributes. Again, I've tested this in my limited environments and it works well for me, but further testing and feedback would be greatly appreciated.  I am also working on a `nomad_deployment` resource as well and will submit that PR once I've tested it a little more.

### DIFF
--- a/nomad/data_source_deployment_test.go
+++ b/nomad/data_source_deployment_test.go
@@ -5,8 +5,8 @@ import (
 	"regexp"
 	"testing"
 
-	"github.com/hashicorp/terraform/helper/resource"
-	"github.com/hashicorp/terraform/terraform"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/terraform"
 )
 
 func TestAccDataSourceDeployment(t *testing.T) {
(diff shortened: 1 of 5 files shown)


nomad/data_source_deployment_test.go
```diff
@@ -0,0 +1,100 @@
...
+						max_file_size = 10
+					}
+				}
+			}
+		}
+	EOT
+}
+
+data "nomad_deployment" "foobaz" {
+	deployment_id			=	""
```
cgbaker: is this intended to be empty?

nomad/data_source_deployment_test.go
```diff
@@ -0,0 +1,100 @@
...
+)
+
+func TestAccDataSourceDeployment(t *testing.T) {
+	resource.Test(t, resource.TestCase{
+		Providers: testProviders,
+		PreCheck:  func() { testAccPreCheck(t) },
+		Steps: []resource.TestStep{
+			{
+				Config: testAccCheckDataSourceNomadDeploymentConfig,
+				Check:  testAccCheckDataSourceNomadDeploymentExist("data.nomad_deployment.foobaz"),
```
cgbaker: this test fails for me: 
```
--- FAIL: TestAccDataSourceDeployment (0.20s)
    testing.go:568: Step 0 error: errors during refresh:

        Error: error checking for deployment: &errors.errorString{s:"Unexpected response code: 500 (missing deployment ID)"}
```

lgfa29:
Hi @slapula, 

I am sorry for the (very) long delay on getting this approved. I fixed the test errors and have also rebased this PR so it's updated with `master`, but we would need you to accept the HashiCorp's CLA.

If you are not OK with that it's totally fine and we will just close the PR without merging it.

Thanks!

PR #39, opened 2018-04-16
Update to use the Nomad 0.8 jobspec.

paddycarver:
Vendor Nomad 0.8's jobspec so that users using Nomad 0.8 don't get false
errors when validating jobs.


paddycarver:
So this started with a `govendor update github.com/hashicorp/nomad/jobspec`. Then that gave build errors. I tried resolving things one at a time, but it became a game of whackamole. I ended up doing `govendor add +external` to try and get everything. It looks like some upstream change requires a package in Nomad that essentially pulls in all of Nomad? Not 100% sure. If you're more comfortable with it, I can try and inventory where these changes came from, but IME, I've generally been pretty unsuccessful at pruning down the number of updated vendor files needed to get the build to work.

AlexeyKhristov:
any estimates when this PR will be merged?

cgbaker:
closing... handled this in #47

PR #52, opened 2019-02-24
Run `terraform fmt` over the example code in the docs

issyl0:
- It's easier on the eyes if everything is lined up, and this makes all
  of the existing docs consistent.


issyl0:
I just rebased this on master, so it's newly mergeable. Any chance of getting this merged?

PR #64, opened 2019-05-17
Monitor job after registration until deployment is complete

### LINKED ISSUE #56: Feature request: Wait for successful job start
```
Terraform v0.11.8
+ provider.nomad v1.3.0
```

One thing I've come to expect from Terraform is that it waits for success on the infrastructure object it creates, as much as is possible.

eg. Terraform waits on creating a AWS Instance, until the instance is running - not just that the AWS API has given a 200 OK back on submitting the HTTP request.

When creating new jobs with the nomad provider it seems like all it looks for is that the API accepted the request - not that the job could be started.

I would like to be able to wait for the job to start, to enable the following use cases.

1. Using the properties of the job that are only available while it's active in downstream objects.
2. Using batch tasks as a form of lifecycle hooks ( before changing service A, run batch B) - as nomad itself seems to be missing this concept, and terraform is a good place for this dependency-graph driven execution.

cgbaker:
Resolves #56 

This PR adds the option for the Nomad TF provider to monitor the deployment on job create/update, until such time as it completes or fails. This requires that a deployment is created, which requires that the job have an `update` stanza.

### DIFF
--- a/nomad/resource_job.go
+++ b/nomad/resource_job.go
@@ -7,6 +7,7 @@ import (
 	"reflect"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/hashicorp/nomad/api"
 	"github.com/hashicorp/terraform/helper/schema"
@@ -45,6 +46,13 @@ func resourceJob() *schema.Resource {
 
 			"deregister_on_id_change": {
 				Description: "If true, the job will be deregistered when the job ID changes.",
+				Optional:    true,
+				Default:     true,
+				Type:        schema.TypeBool,
+			},
+
+			"detach": {
+				Description: "If true, the provider will return immediately after creating or updating, instead of monitoring.",
 				Optional:    true,
 				Default:     true,
 				Type:        schema.TypeBool,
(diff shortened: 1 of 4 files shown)


nomad/resource_job.go
```diff
@@ -169,27 +195,133 @@ func resourceJobRegister(d *schema.ResourceData, meta interface{}) error {
...
+
+	// deploymentId is the deployment that we are monitoring. This is captured from the
+	// final evaluation.
+	deploymentId string
+}
+
+// deploymentStateRefreshFunc returns a resource.StateRefreshFunc that is used to watch
+// the deployment from a job create/update
+func deploymentStateRefreshFunc(client *api.Client, initialEvalId string) resource.StateRefreshFunc {
+	watching := &monitorState{
```
notnoop: The `monitorState` struct seems only used here.  What's its benefit compared to having `evalId` and `deploymentId` variables here without the struct?

nomad/resource_job.go
```diff
@@ -50,6 +52,25 @@ func resourceJob() *schema.Resource {
 				Type:        schema.TypeBool,
 			},
 
+			"detach": {
```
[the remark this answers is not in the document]
cgbaker: i originally had it as `monitor`, but i thought that nomad users (especially those converting `nomad` CLI commands to the TF provider) might be more comfortable with the Nomad nomenclature.

nomad/resource_job.go
cgbaker: i had originally (and incorrectly) thought i could have terraform pass it in, so i had it as a struct. but i was wrong, and I moved it into this function factory. 

this can be exploded it into multiple variables, but the risk there is that someone adding a new tracked data must remember to make that data a pointer so that it persists across calls to the refresh function, whereas I've taken care of that.

PR #67, opened 2019-06-06
go mod tidy

radeksimko:


cgbaker:
which version of go are you using? my `go mod tidy` with go 1.11.9 doesn't see any changes.

radeksimko:
@cgbaker I'm using `1.11.5` per goenv configuration in the root. To be more specific I get the following error when running `go mod tidy`:

```
go: verifying github.com/hashicorp/go-getter@v1.1.0: checksum mismatch
	downloaded: h1:iGVeg7L4V5FTFV3D6w+1NAyvth7BIWWSzD60pWloe2Q=
	go.sum:     h1:fb6yisdpoNLiQZsuZqafiuaSSXFjxByBm/YS3k8MXjo=
```
which is why I `rm go.sum` and ran `go mod tidy` which generated `go.sum` from scratch.

I'm able to reproduce this also in a clean `golang:1.11.5` Docker container.

I assume this could be version difference and/or corrupted module cache.

cgbaker:
so, if I delete `go.sum` and re-run, i get different answers in a clean `golang:1.11.9` container (`go version go1.11.9 linux/amd64`) than i do directly on my laptop (`go version go1.11.9 darwin/amd64`)

i wonder if it's another checksum bug like fixed in 1.11.4. i'll go ahead and merge this.

PR #68, opened 2019-06-14
Add support for nomad 0.9.3 update stanza "auto_promote" property

transferkraM:


transferkraM:
* https://www.nomadproject.io/docs/job-specification/update.html#auto_promote
* https://github.com/hashicorp/nomad/pull/5719

cgbaker:
Thanks for this, will look at it when I get back to work next week.

JanBerktold:
@cgbaker What is the status here? :)

nomad/resource_job_test.go
```diff
@@ -885,6 +886,7 @@ resource "nomad_job" "test" {
 				healthy_deadline = "6m"
 				progress_deadline = "11m"
 				auto_revert = true
+        auto_promote = true
```
cgbaker: this test was looking at new config for the 0.9.0 API... `auto_promote` should be part of a separate test for 0.9.3. such a test should verify that this option is successfully being submitted to Nomad.

nomad/jobspec/parse.go
```diff
@@ -1603,6 +1603,7 @@ func parseUpdate(result **api.UpdateStrategy, list *ast.ObjectList) error {
 		"healthy_deadline",
 		"progress_deadline",
 		"auto_revert",
+		"auto_promote",
```
cgbaker: the `api.UpdateStrategy` object that this attempting to parse into doesn't have an `AutoPromote` field, because the dependency version for the Nomad API needs to be updated.

nomad/datasource_nomad_job.go
```diff
@@ -242,6 +242,10 @@ func dataSourceJob() *schema.Resource {
 										Type:     schema.TypeBool,
 										Computed: true,
 									},
+									"auto_promote": {
+										Type:     schema.TypeBool,
+										Computed: true,
+									},
```
cgbaker: please add a test to `datasource_nomad_job_test.go` to test this is working; i don't believe that it is.

AndrewScibek:
I made a pr against his fork that makes this functionality work for my use case(updating the dependencies). https://github.com/transferkraM/terraform-provider-nomad/pull/1

@cgbaker can you expand how i can make the test for different versions or point me to the docs that let me know how I can do it? This is a useful feature that i would like merged

cgbaker:
Thanks @AndrewScibek. I'll take a look at this.

mikeblum:
Any update on this PR? Looking forward to using this.

PR #73, opened 2019-08-30
Fix delete default namespace

### LINKED ISSUE #72: Provider should not attempt to delete `default` namespace
Referencing the `default` namespace (e.g., while attaching a quota) will cause the TF provider to attempt to delete the namespace on `destroy`. This results in a 500 error. There are workarounds for this, but the root issue is that the TF provider should not attempt to delete things that we know _a priori_ cannot be deleted.

lgfa29:
Currently if you reference the `default` namespace in your configuration Terraform will fail to destroy it because the API will return a 500 error:

```
An execution plan has been generated and is shown below.
Resource actions are indicated with the following symbols:
  - destroy

  # nomad_namespace.default will be destroyed
  - resource "nomad_namespace" "default" {
      - description = "Default shared namespace" -> null
      - id          = "default" -> null
      - name        = "default" -> null
    }

Plan: 0 to add, 0 to change, 1 to destroy.

Do you want to perform these actions?
  Terraform will perform the actions described above.
  Only 'yes' will be accepted to approve.

  Enter a value: yes

nomad_namespace.default: Destroying... [id=default]

Error: error deleting namespace "default": Unexpected response code: 500 (can not delete default namespace)
```

This error can prevent other resources from being destroyed if they depend on this namespace.

This PR prevents this error from happening by not trying to delete the `default` namespace. Instead it resets the `quota` and `description` attributes to their initial values.

Fixes #72 .

### DIFF
--- a/nomad/resource_namespace_test.go
+++ b/nomad/resource_namespace_test.go
@@ -6,6 +6,7 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/hashicorp/nomad/api"
 	"github.com/hashicorp/terraform/helper/acctest"
 	"github.com/hashicorp/terraform/helper/resource"
 	"github.com/hashicorp/terraform/terraform"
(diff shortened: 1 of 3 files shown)


nomad/resource_namespace_test.go
```diff
@@ -187,6 +204,26 @@ func testResourceNamespace_checkDestroy(name string) resource.TestCheckFunc {
...
+			return fmt.Errorf("failed to find default namespace %q.", defaultNamespace.Name)
+		}
+		if namespace.Description != defaultNamespace.Description || namespace.Quota != defaultNamespace.Quota {
+			return fmt.Errorf("default namespace %q not reset.", defaultNamespace.Name)
+		}
+
+		return nil
+	}
+}
+
```
cgbaker: 👍

website/docs/r/namespace.html.markdown
```diff
@@ -13,6 +13,11 @@ Provisions a namespace within a Nomad cluster.
 ~> **Enterprise Only!** This API endpoint and functionality only exists in
 Nomad Enterprise. This is not present in the open source version of Nomad.
 
+Nomad auto-generates a default namespace called `default`. This namespace
+cannot be removed, so destroying a `nomad_namespace` resource where
+`name = "default"` will cause the namespace to be reset to its initial
```
cgbaker:
```suggestion
`name = "default"` will cause the namespace to be reset to its default
```

PR #75, opened 2019-09-11
remove use of config pkg

appilon:
In preparation for upcoming standalone SDK, this PR removes the usage of removed functions/packages. Upgrades terraform to v0.12.8


appilon:
no new failures in TC!

PR #88, opened 2019-10-19
add data source for nomad ACL tokens

### LINKED ISSUE #85: Create data source for ACL token
# nomad_acl_token

Get information on an [ACL token](https://www.nomadproject.io/api/acl-tokens.html#read-token). 

**Important:** this data source will store tokens in the Terraform state. The documentation should have a warning like the [one in the resource page](https://github.com/terraform-providers/terraform-provider-nomad/blob/master/website/docs/r/acl_token.html.markdown).

## Example usage

```hcl
data "nomad_acl_token" "my_token" {
  accessor_id = "aa534e09-6a07-0a45-2295-a7f77063d429"
}
```

## Argument reference

* `accessor_id`: `(string)` Non-sensitive identifier for this token.

## Attributes reference

* `name`: `(string)` Human-friendly name of the ACL token.
* `accessor_id`: `(string)` Non-sensitive identifier for this token.
* `secret_id`: `(string)` The token value itself.
* `type`: `(string)` The type of the token.
* `policies`: `(list of strings)` List of policy names associated with this token.
* `global`: `(bool)` Whether the token is replicated to all regions, or if it will only be used in the region it was created.
* `create_time`: `(string)` Date and time the token was created.

softpunks:
This PR adds the data source for ACL tokens resource.

closes #85

### DIFF
--- a/nomad/data_source_acl_token.go
+++ b/nomad/data_source_acl_token.go
@@ -86,7 +86,7 @@ func dataSourceACLTokenRead(d *schema.ResourceData, meta interface{}) error {
 	d.Set("policies", token.Policies)
 	d.Set("secret_id", token.SecretID)
 	d.Set("global", token.Global)
-	d.Set("create_time", token.CreateTime)
+	d.Set("create_time", token.CreateTime.UTC().String())
 
 	return nil
 }
--- a/nomad/data_source_acl_token.go
+++ b/nomad/data_source_acl_token.go
@@ -73,7 +73,7 @@ func dataSourceACLTokenRead(d *schema.ResourceData, meta interface{}) error {
 		// rather than a nil result, so we must check this way.
 		if strings.Contains(err.Error(), "404") {
 			d.SetId("")
-			return err
+			return nil
 		}
 
 		return fmt.Errorf("error reading ACL token %q: %s", accessor, err.Error())
--- a/nomad/data_source_acl_token.go
+++ b/nomad/data_source_acl_token.go
@@ -1,0 +1,105 @@
+package nomad
+
+import (
+	"fmt"
+	"log"
+	"strings"
+
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+)
+
+func dataSourceACLToken() *schema.Resource {
+	return &schema.Resource{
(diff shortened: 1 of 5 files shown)


website/docs/d/acl_token.html.markdown
```diff
@@ -0,0 +1,40 @@
...
+layout: "nomad"
+page_title: "Nomad: nomad_acl_token"
+sidebar_current: "docs-nomad-datasource-acl-token"
+description: |-
+  Get information on an ACL token.
+---
+
+# nomad_acl_token
+
+Get information on an ACL token
```
lgfa29: Add a period at the end of the sentence.

website/nomad.erb
```diff
@@ -25,6 +25,9 @@
             <li<%= sidebar_current("docs-nomad-datasource-regions") %>>
               <a href="/docs/providers/nomad/d/regions.html">nomad_regions</a>
             </li>
+            <li<%= sidebar_current("docs-nomad-datasource-acl-token") %>>
+              <a href="/docs/providers/nomad/d/acl_token.html">nomad_acl_token</a>
+          </li>
```
lgfa29: Move this item to the top of the list to keep it alphabetically sorted.

nomad/data_source_acl_token.go
```diff
@@ -0,0 +1,92 @@
...
+	// retrieve the token
+	log.Printf("[DEBUG] Reading ACL Token %q", accessor)
+	token, _, err := client.ACLTokens().Info(accessor, nil)
+	if err != nil {
+
+		// As of Nomad 0.4.1, the API client returns an error for 404
+		// rather than a nil result, so we must check this way.
+		if strings.Contains(err.Error(), "404") {
+			d.SetId("")
+			return err
```
lgfa29: You don't need to return `err` in this case, you can just `return nil` since this is not a failure.

softpunks:
@cgbaker I added the test for `create_time` but I can't seem to get it to pass. Not sure if the problem is a) a bug in my test or b) `create_time` isn't getting set in the course of the test for some reason. Could anyone point me in the right direction on this?

Issue #82, opened 2019-10-15
Create data source for ACL policy


lgfa29:
# nomad_acl_policy

Get information on an [ACL policy](https://www.nomadproject.io/api/acl-policies.html#read-policy). 

## Example usage

```hcl
data "nomad_acl_policy" "my_policy" {
  name = "my-policy"
}
```

## Argument reference

* `name`: `(string)` Name of the ACL policy.

## Attributes reference

* `name`: `(string)` Name of the ACL policy.
* `description`: `(string)` ACL policy description.
* `rules`: `(string)` Policy rules in HCL or JSON format.

ddreier:
I believe that I have this mostly done. Will submit a PR soon.

cgbaker:
closed by #89

PR #89, opened 2019-10-20
Add Data Source for a Nomad ACL Policy.

ddreier:
For #82.

Needed for the Test to work:
 * Instance of Nomad with ACL enabled
 * Run the `nomad acl bootstrap` command to get a secret_id
 * Set the NOMAD_TOKEN environment variable to the secret_id

Given that there are already tests for the `nomad_acl_policy` resource that actually create and delete policies, I don't think that this should be a problem.


ddreier:
Thank you for the feedback! I'm working on getting it all implemented now.

ddreier:
I believe that I have everything addressed. Let me know if there's anything else. Thanks!

lgfa29:
Thanks for the work @ddreier, just a couple of things left that I noticed:
* add a [menu item](https://github.com/terraform-providers/terraform-provider-nomad/blob/master/website/nomad.erb#L15) for the new data source
* the file name and test function name don't need to have `nomad` in them

lgfa29:
I also just merged a PR that might conflict with your work, but it should be simple to fix it. Let me know if you need any help on that.

ddreier:
@lgfa29 thanks! I have pushed those changes.

PR #103, opened 2020-05-15
nomad api 0.11.2 support

cgbaker:


cgbaker:
I'd appreciate that, thank you

PR #111, opened 2020-05-28
Fix panic when using invalid JSON jobspec

lgfa29:
Currently, when the input `jobspec` is in JSON format, but the content doesn't follow the expected Nomad JSON structure, the provider will panic when trying to call the Nomad API.

This PR changes the jobpsec parsing logic to gracefully fail when the in this situation. 

It also allows for the JSON input to have a root element so that the provider can handle outputs from the Nomad CLI directly:

```shellsession
$ nomad job run -output jobspec.hcl
{
    "Job": {
        "Stop": null,
        "Region": "us",
...
}
```

### DIFF
--- a/nomad/resource_job.go
+++ b/nomad/resource_job.go
@@ -548,21 +548,50 @@ func resourceJobCustomizeDiff(d *schema.ResourceDiff, meta interface{}) error {
 
 func parseJobspec(raw string, is_json bool, vaultToken *string) (*api.Job, error) {
 	var job *api.Job
+	var err error
+
 	if is_json {
-		err := json.Unmarshal([]byte(raw), &job)
-		if err != nil {
-			return nil, fmt.Errorf("error parsing jobspec: %s", err)
-		}
+		job, err = parseJSONJobspec(raw)
 	} else {
-		var err error
 		job, err = jobspec.Parse(strings.NewReader(raw))
-		if err != nil {
-			return nil, fmt.Errorf("error parsing jobspec: %s", err)
-		}
+	}
+	if err != nil {
+		return nil, fmt.Errorf("error parsing jobspec: %s", err)
+	}
+
+	// If job is empty after parsing, the input is not a valid Nomad job.
+	if reflect.DeepEqual(job, &api.Job{}) {
+		return nil, fmt.Errorf("error parsing jobspec: input JSON is not a valid Nomad jobspec")
 	}
 
 	// Inject the Vault token
 	job.VaultToken = vaultToken
+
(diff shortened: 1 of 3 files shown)


nomad/resource_job.go
```diff
@@ -548,17 +548,20 @@ func resourceJobCustomizeDiff(d *schema.ResourceDiff, meta interface{}) error {
...
-		var err error
 		job, err = jobspec.Parse(strings.NewReader(raw))
-		if err != nil {
-			return nil, fmt.Errorf("error parsing jobspec: %s", err)
-		}
+	}
+	if err != nil {
+		return nil, fmt.Errorf("error parsing jobspec: %s", err)
+	}
+
```
jrasell, line 561: The job is a pointer, do we also need to perform a nil check here before moving on or will we always get an error if the job is nil?

cgbaker: `reflect.DeepEqual` below will handle the nil and empty check.

jrasell: ah yeah, duh :D

nomad/resource_job.go
```diff
@@ -567,6 +570,32 @@ func parseJobspec(raw string, is_json bool, vaultToken *string) (*api.Job, error
...
+	jobBytes, ok := root["Job"]
+	if !ok {
+		// Parse the input as is if there's no "Job" root.
+		jobBytes = []byte(raw)
+	}
+
+	// Parse actual job.
+	var job *api.Job
+	err = json.Unmarshal(jobBytes, &job)
```
cgbaker, line 591: is it necessary to pass a pointer-to-pointer to `Unmarshall` if we don't care about the null case? or necessary at all? this call is so weird.

nomad/resource_job.go
cgbaker: https://github.com/golang/go/blob/master/src/reflect/deepequal.go#L190-L192

cgbaker: `&api.Job` is not `nil`, but if `job == nil`, then `DeepEqual` appears to return `nil == &api.Job{}`, which is false. you are correct, i had the logic reversed here.

nomad/resource_job.go
lgfa29: I copied this as it was, but now that you mentioned it does look a bit weird. The pointer-to-pointer is necessary here because otherwise we would be passing `nil` to `json.Unmarshal`.

I think a better version would be:

```go
job := &api.Job{}
err = json.Unmarshal(jobBytes, &job)
```

PR #128, opened 2020-07-11
Add nomad_acl_tokens data source

remilapeyre:
Closes https://github.com/terraform-providers/terraform-provider-nomad/issues/86

### DIFF
--- a/nomad/data_source_acl_tokens.go
+++ b/nomad/data_source_acl_tokens.go
@@ -1,0 +1,83 @@
+package nomad
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+)
+
+func dataSourceACLTokens() *schema.Resource {
+	return &schema.Resource{
+		Read: aclTokensDataSourceRead,
+
+		Schema: map[string]*schema.Schema{
+			"prefix": {
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+
+			"acl_tokens": {
+				Type:     schema.TypeList,
+				Computed: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"accessor_id": {
+							Type:     schema.TypeString,
+							Computed: true,
+						},
+						"name": {
+							Type:     schema.TypeString,
+							Computed: true,
+						},
+						"type": {
+							Type:     schema.TypeString,
+							Computed: true,
+						},
+						"policies": {
+							Type:     schema.TypeList,
+							Computed: true,
+							Elem:     &schema.Schema{Type: schema.TypeString},
+						},
+						"global": {
+							Type:     schema.TypeBool,
+							Computed: true,
+						},
+						"create_time": {
+							Type:     schema.TypeString,
+							Computed: true,
+						},
+					},
+				},
+			},
+		},
+	}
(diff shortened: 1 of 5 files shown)


nomad/data_source_acl_tokens.go
```diff
@@ -0,0 +1,83 @@
...
+				},
+			},
+		},
+	}
+}
+
+func aclTokensDataSourceRead(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(ProviderConfig).client
+
+	tokens, _, err := client.ACLTokens().List(nil)
```
cgbaker, line 60: `List()` can be passed a `QueryOptions` including a prefix:
https://github.com/hashicorp/nomad/blob/dc6aa33916faf4e6129e138f483e0e6b88c975b7/api/acl.go#L86
https://github.com/hashicorp/nomad/blob/dc6aa33916faf4e6129e138f483e0e6b88c975b7/api/api.go#L59

This will reduce the size of the response and avoid the need for filtering below.

website/nomad.erb
```diff
@@ -19,6 +19,9 @@
             <li<%= sidebar_current("docs-nomad-datasource-acl-token") %>>
               <a href="/docs/providers/nomad/d/acl_token.html">nomad_acl_token</a>
             </li>
+            <li<%= sidebar_current("docs-nomad-datasource-acl-tokens") %>>
+              <a href="/docs/providers/nomad/d/acl_token.html">nomad_acl_tokens</a>
```
cgbaker, line 23:
```suggestion
              <a href="/docs/providers/nomad/d/acl_tokens.html">nomad_acl_tokens</a>
```

nomad/data_source_acl_tokens.go
remilapeyre: Thanks!

nomad/data_source_acl_tokens.go
```diff
@@ -0,0 +1,82 @@
...
+			"accessor_id": t.AccessorID,
+			"name":        t.Name,
+			"type":        t.Type,
+			"policies":    t.Policies,
+			"global":      t.Global,
+			"create_time": t.CreateTime.String(),
+		}
+	}
+
+	d.SetId("nomad-tokens")
```
cgbaker, line 80: apologies for not noticing this before... does this ID need to include the prefix? different `data "nomad_acl_tokens"` with different prefix options should (probably?) have different IDs.

cgbaker: i manually tested an terraform didn't complain... i will ask someone from the provider team on Monday.

remilapeyre: I think the ID does not matters as long as it is not `""` as this is used to indicate, it's usually set to the external ID for resources that have one but when a Terraform resource is a virtual resource (like https://github.com/hashicorp/terraform-provider-null/blob/master/null/data_source.go#L51) or is actually multiple resource it can be set to something else.

I can push a commit to change it to `d.SetId(client.Address() + "/tokens/" + prefix)` which would make it more like the pattern used by the other datasources if you want.

cgbaker:
yeah, I talked to some folks on the plugins team and they confirmed the
value doesn't matter. you can merge it as is.

cgbaker:
thanks, @remilapeyre!

PR #131, opened 2020-07-13
install and use nomad for testing

cgbaker:
now that enterprise binaries are freely available for short-term use, we should deploy nomad in travis for testing the provider.

this WIP still needs
* `before_script` stage: to run `scripts/start-nomad.sh`
* `after_script` stage: to run `scripts/stop-nomad.sh`
* another `install` to fetch/install vault


scripts/getvault.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+VAULT_BINARY=https://releases.hashicorp.com/vault/vault_1.5.0+ent/vault_1.5.0+ent_linux_amd64.zip
```
cgbaker, line 3:
```suggestion
VAULT_BINARY=https://releases.hashicorp.com/vault/1.5.0+ent/vault_1.5.0+ent_linux_amd64.zip
```

PR #149, opened 2020-08-31
Add a nomad_job_v2 resource

remilapeyre:


nomad/resource_job_fields.go
```diff
@@ -0,0 +1,1081 @@
...
+		"consul_token": {
+			Type:     schema.TypeString,
+			Optional: true,
+		},
+		"constraint": getConstraintFields(),
+		"affinity":   getAffinityFields(),
+		"spread":     getSpreadFields(),
+		"group":      getGroupFields(),
+		"parameterized": {
+			Type:     schema.TypeList,
```
cgbaker, line 58: why list? this feels like some Terraform clever secret that I would like to learn 😄

nomad/resource_job_fields.go
```diff
@@ -0,0 +1,1081 @@
...
+				"mode": {
+					Type:     schema.TypeString,
+					Optional: true,
+				},
+				"port": {
+					Type:     schema.TypeList,
+					Optional: true,
+					MaxItems: 1,
+					Elem: &schema.Resource{
+						Schema: map[string]*schema.Schema{
```
cgbaker, line 986: this will need a `label` field as well... that is the name of the port resource.

nomad/resource_job_v2.go
```diff
@@ -0,0 +1,1682 @@
...
+			if err != nil {
+				return nil, err
+			}
+			template.Splay = duration
+
+			duration, err = getDuration(tp["vault_grace"])
+			if err != nil {
+				return nil, err
+			}
+			template.VaultGrace = duration
```
lgfa29, line 625: I think it would be better to avoid re-using the `duration` variable here. It could lead to bugs in the future.

nomad/resource_job_v2.go
```diff
@@ -0,0 +1,1682 @@
...
+		}
+		task.KillTimeout = duration
+
+		duration, err = getDuration(t["shutdown_delay"])
+		if err != nil {
+			return nil, err
+		}
+		if duration != nil {
+			task.ShutdownDelay = *duration
+		}
```
lgfa29, line 689: Same issue here on re-using the `duration` variable.

nomad/resource_job_v2.go
```diff
@@ -0,0 +1,1682 @@
...
+	return tasks, nil
+}
+
+func getNetworks(d interface{}) []*api.NetworkResource {
+	networks := make([]*api.NetworkResource, 0)
+
+	for _, nr := range d.([]interface{}) {
+		n := nr.(map[string]interface{})
+
+		// TODO(remi): find what to do with the ports here
```
lgfa29, line 703: The [`parsePorts` function checks if there's a Value set](https://github.com/hashicorp/nomad/blob/master/jobspec/parse_network.go#L105-L109) (where [`Value` maps to `static`](https://github.com/hashicorp/nomad/blob/master/api/resources.go#L88) in the jobspec).

I think that's what you would need to do here as well.

nomad/resource_job_v2.go
```diff
@@ -0,0 +1,1682 @@
...
+}
+
+func getVault(d interface{}) *api.Vault {
+	for _, vlt := range d.([]interface{}) {
+		v := vlt.(map[string]interface{})
+
+		policies := make([]string, 0)
+		for _, p := range v["policies"].([]interface{}) {
+			policies = append(policies, p.(string))
+		}
```
lgfa29, line 998: Would it make sense to encapsulate the logic f parsing a `[]interface{}` as `[]string` into a function, like `getMapOfString`? This seems to be used in a lot of places.

nomad/resource_job_v2.go
```diff
@@ -0,0 +1,1682 @@
...
+	}
+
+	return res
+}
+
+func readGroups(d *schema.ResourceData, groups []*api.TaskGroup) (interface{}, error) {
+	res := make([]interface{}, 0)
+
+	// we have to look for the groups the user created in its configuration as
+	// we will need to set the "ephemeral_disk" and "restart" block only if he
```
lgfa29, line 1078:
```suggestion
	// we will need to set the "ephemeral_disk" and "restart" block only if they
```

nomad/resource_job_v2.go
```diff
@@ -0,0 +1,1682 @@
...
+					"interval": "24h0m0s",
+					"mode":     "fail",
+				}
+			} else if _type == "system" {
+				defaultValue = map[string]interface{}{
+					"attempts": 2,
+					"delay":    "15s",
+					"interval": "30m0s",
+					"mode":     "fail",
+				}
```
lgfa29, line 1140: Does Nomad require these values to be set? If possible, it would be better to submit the job without them and let Nomad set their defaults.

nomad/resource_job_v2.go
```diff
@@ -0,0 +1,1682 @@
...
+			return nil, err
+		}
+
+		isSet := len(currentConfig["reschedule"].([]interface{})) > 0
+		reschedule, err := readReschedule(d, isSet, g.ReschedulePolicy)
+		if err != nil {
+			return nil, err
+		}
+
+		isSet = len(currentConfig["migrate"].([]interface{})) > 0
```
lgfa29, line 1180: Same as before about reusing `isSet` here.

nomad/resource_job_v2.go
```diff
@@ -0,0 +1,1682 @@
...
+	// }
+	if err = d.Set("constraint", readConstraints(job.Constraints)); err != nil {
+		return fmt.Errorf("Failed to set 'constraint': %v", err)
+	}
+	if err = d.Set("affinity", readAffinities(job.Affinities)); err != nil {
+		return fmt.Errorf("Failed to set 'affinity': %v", err)
+	}
+	if err = d.Set("spread", readSpreads(job.Spreads)); err != nil {
+		return fmt.Errorf("Failed to set 'spread': %v", err)
+	}
```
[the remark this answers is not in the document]
remilapeyre: We had a few bugs in the Consul provider where we either had a type in the attribute we wanted to set, were using the wrong type, or were doing something completely unsupported (e.g. `d.Set("attr.0.foo", ...)`) and adding those checks showed some of those bugs.

I like `stateWriter` a lot thought as the verbosity always bothered me, it's a great solution :) I stole it for the Consul provider too.

nomad/resource_job_fields.go
remilapeyre: As the issue @lgfa29 linked explains you can't have a complex resource in a `schema.TypeMap` attribute. If you try to do it, it may look like it's working but all attributes get their type converted. This is caught by https://github.com/bflad/tfproviderlint.

nomad/resource_job_fields.go
```diff
@@ -0,0 +1,1081 @@
...
+func getJobFields() map[string]*schema.Schema {
+	return map[string]*schema.Schema{
+		"namespace": {
+			Type:     schema.TypeString,
+			Default:  "default",
+			Optional: true,
+		},
+		"priority": {
+			Type:     schema.TypeInt,
+			Default:  50,
```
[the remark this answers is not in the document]
remilapeyre: The Nomad API does not require a default for `priority`, if we don't set one a default will be used automatically so we could set `Computed: true` here. The issue when doing that is that if an operator later manually change the `priority` value to something else, 75 for example, Terraform will be unable to detect that this change should be tracked and will not report a diff as the attribute is set as `Computed`.

Does this make sense?

nomad/resource_job_fields.go
```diff
@@ -0,0 +1,1081 @@
...
+				"progress_deadline": {
+					Type:     schema.TypeString,
+					Default:  "0s",
+					Optional: true,
+				},
+				"stagger": {
+					Type:     schema.TypeString,
+					Default:  "0s",
+					Optional: true,
+				},
```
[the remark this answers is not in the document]
remilapeyre: Yes, I have not look for what Nomad uses as defaults here. I think we may be obligated to use the same the defaults as the Nomad API in the Terraform provider for the reasons I gave in https://github.com/hashicorp/terraform-provider-nomad/pull/149#discussion_r487458932 thought.

nomad/resource_job_v2.go
remilapeyre: This is actually linked to https://github.com/hashicorp/terraform-plugin-sdk/issues/62. If the user did not set a `restart policy it is automatically set by Nomad and returned in the job spec. In this case Terraform gets a list of one element but does not expect any a perpetually creates a diff for this resource.

We had this issue in the Consul provider (https://github.com/hashicorp/terraform-provider-consul/pull/121 and https://github.com/hashicorp/terraform-provider-consul/issues/119) and the only solution I could find was to check whether the user had set the attribute and if the server had sent back the default values for this attribute and ignore it in this case. It's cumbersome but I'm not aware of a better solution at the moment. Does this make sense?

nomad/resource_job_v2.go
remilapeyre: Thanks!

nomad/resource_job_fields.go
```diff
@@ -0,0 +1,1081 @@
...
+			Optional: true,
+		},
+		"type": {
+			Type:     schema.TypeString,
+			Default:  "service",
+			Optional: true,
+		},
+		"region": {
+			Type:     schema.TypeString,
+			Default:  "global",
```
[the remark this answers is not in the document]
remilapeyre: In theory, this should be fixed by 3bb15606299dd7a5a411be98590f643d1b52800f, I still need to write tests for it thought.

remilapeyre:
There is one thing were I'm not sure, the documentation for the spread target at https://www.nomadproject.io/docs/job-specification/spread#spread-with-target-percentages give them an ID:

```hcl
spread {
  attribute = "${node.datacenter}"
  weight    = 100

  target "us-east1" {
    percent = 50
  }
}
```

but the struct in the API at https://github.com/hashicorp/nomad/blob/master/api/tasks.go#L263-L267 has no attribute where to set it. Is it useless?

lgfa29:
Despite @remilapeyre brave efforts, this approach turned out to be impossible to manage over time as it requires keeping the Terraform spec for jobs in sync with Nomad. 

We should investigate the advancements in the [Terraform plugin development ecosystem](https://developer.hashicorp.com/terraform/plugin/terraform-plugin-protocol) to come up with a better solution to #1.

Thank you again @remilapeyre for all the hard work on this!

PR #157, opened 2020-10-23
Add Scheduler config resource

### LINKED ISSUE #153: New Resource - Operator Scheduler Configurations
It would be fantastic if we could get support for managing the new Scheduler Configuration. I'm specifically interested in managing SchedulerAlgorithm=spread for example. Here's a link to the api section describing this:
https://www.nomadproject.io/api-docs/operator#scheduleralgorithm-1

daktari:
```
resource "nomad_scheduler_config" "cfg" {
algorithm = "binpack"
        preemption {
            system_enabled = true
            batch_enabled = false
            service_enabled = false
        }
}
```

closes #153


jrasell:
Hi @daktari and thanks for raising this PR. It seems the `TestSchedulerConfig_basic` you add is currently failing as detailed above and within Travis. Is this something you could look into fixing?

```
=== RUN   TestSchedulerConfig_basic
    testing.go:684: Step 0 error: config is invalid: "preemption": required field is not set
```

lgfa29 (approved):

PR #160, opened 2020-10-29
Use jobspec package from nomad directly

notnoop:
Also update some of the dependencies


go.mod
```diff
@@ -11,14 +11,18 @@ exclude (
 	github.com/Sirupsen/logrus v1.4.1
 )
 
+// Force using go-getter version from terraform plugin
+replace github.com/hashicorp/go-getter => github.com/hashicorp/go-getter v1.4.0
```
notnoop, line 15: This is a hack :( - will not be needed once jobspec is an explicit module.

notnoop:
superceeded by https://github.com/hashicorp/terraform-provider-nomad/pull/161

Issue #150, opened 2020-09-02
deps: provider should avoid importing `github.com/hashicorp/vault`


jrasell:
Vault issue comment reference: https://github.com/hashicorp/vault/issues/9575#issuecomment-665013482

The Nomad provider currently imports `github.com/hashicorp/vault` because of its use of `config.DefaultTokenHelper()` within `provider.go`. A per the above discussion Vault imports should be limited to `sdk` and `api` otherwise problems may occur.

A quick look indicates there is no token helper functionality within either of the recommended packages, although this was not exhaustive.

tristanmorgan:
A note for anyone caught when you broke your Vault token helper.... this is the error message displayed.
```
╷
│ Error: Error getting token helper: unknown error getting the external helper path
│ 
│   with provider["registry.terraform.io/hashicorp/nomad"],
│   on <input-prompt> line 1:
│   (source code not available)
│ 
╵
```
Is there a reason why a vault misconfiguration should affect a 3 line nomad config?
```hcl
data "nomad_job" "example" {
  job_id = "example"
}
```

lgfa29:
Hi @tristanmorgan,

The Nomad provider uses the same logic as Vault to find a token, so if your Vault token helper is incorrect, the Nomad provider will fail as well.

lgfa29:
Closed by #445

PR #161, opened 2020-10-29
pull jobspec from nomad repo

cgbaker:
require a replace directive because nomad was pulling in a go-getter (which we didn't use) that was incompatible with terraform-plugin-test (which we do use)


go.mod
```diff
@@ -12,13 +12,22 @@ exclude (
...
-	github.com/stretchr/testify v1.5.1
+	github.com/mitchellh/mapstructure v1.3.1 // indirect
+	github.com/onsi/ginkgo v1.12.0 // indirect
+	github.com/stretchr/testify v1.6.1
+	golang.org/x/build v0.0.0-20190111050920-041ab4dc3f9d // indirect
+	golang.org/x/oauth2 v0.0.0-20200107190931-bf48bf16ab8d // indirect
+)
+
+replace (
+    github.com/hashicorp/go-getter => github.com/hashicorp/go-getter v1.4.0
```
notnoop, line 32: I'd suggest adding a comment on why this is needed?

go.mod
```diff
@@ -12,13 +12,22 @@ exclude (
...
+	github.com/google/go-cmp v0.5.0
+	github.com/hashicorp/go-multierror v1.1.0 // indirect
+	github.com/hashicorp/go-version v1.2.1
+	github.com/hashicorp/hcl v1.0.1-0.20201016140508-a07e7d50bbee // indirect
+	github.com/hashicorp/nomad v0.12.5-0.20201029140339-d6255129a300
 	github.com/hashicorp/nomad/api v0.0.0-20201028165800-38e23b62a770
-	github.com/hashicorp/terraform-plugin-sdk v1.15.0
+	github.com/hashicorp/terraform-plugin-sdk v1.16.0
+	github.com/hashicorp/terraform-plugin-test v1.4.3 // indirect
 	github.com/hashicorp/vault v0.10.4
```
notnoop, line 23: wow - that's so old!

notnoop (approved):

go.mod
lgfa29: Old and wrong 😬 
https://github.com/hashicorp/terraform-provider-nomad/issues/150

lgfa29 (dismissed):

lgfa29 (approved):

PR #162, opened 2020-10-30
Add data sources for scaling policies

lgfa29:
Add two new data sources to load scaling policies from Nomad.

Changelog update will come in a separate PR.

#### Usage example:

```hcl
resource "nomad_job" "job" {
  jobspec = <<EOF
job "foo-scaling-policy" {
  datacenters = ["dc1"]
  group "foo" {

    scaling {
      enabled = false
      min     = 1
      max     = 20
          type    = "horizontal"

      policy {
        cooldown = "20s"
      }
    }

    task "foo" {
      driver = "raw_exec"
      config {
        command = "/bin/sleep"
        args    = ["10"]
      }
    }
  }
}
EOF
}

data "nomad_scaling_policies" "policies" {
  job_id = nomad_job.job.name
}

data "nomad_scaling_policy" "policy" {
  id = data.nomad_scaling_policies.policies.policies[0].id
}
```

This PR also includes a new helper used to check for errors when calling `d.Set` in a less verbose way:

```go
sw := helper.NewStateWriter(d)

sw.Set("enabled", p.Enabled)
sw.Set("type", p.Type)
sw.Set("min", p.Min)
sw.Set("max", p.Max)
sw.Set("target", p.Target)
                                                                        
if sw.Error() != nil {
	return sw.Error()
}
```


cgbaker (approved):
LGTM

cgbaker (approved):

PR #165, opened 2020-10-30
Add data source for listing datacenters

### LINKED ISSUE #151: Feature idea: nomad_datacenter data source
Similar to the nomad_namespaces that provides a list of known nomad namespaces, it would be nice to have an equivalent data source that could list the datacenters in a given cluster.  This would make re-usable terraform modules much easier and more automatic.

lgfa29:
This PR introduces a new data source to list datacenters in the region where the provider is configured.

```hcl
data "nomad_datacenters" "datacenters" {
  prefix            = "prod"
  ignore_down_nodes = true
}
```

An [alias](https://www.terraform.io/docs/configuration/providers.html#alias-multiple-provider-configurations) can be used for listing datacenters in multiple regions.

Closes #151

### DIFF
--- a/nomad/data_source_datacenters.go
+++ b/nomad/data_source_datacenters.go
@@ -1,0 +1,80 @@
+package nomad
+
+import (
+	"fmt"
+	"sort"
+	"strings"
+
+	"github.com/hashicorp/nomad/api"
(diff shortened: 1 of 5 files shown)


cgbaker (approved):
nice

PR #166, opened 2020-10-30
update terraform SDK to 2.4.0 and Nomad to 1.0.2

cgbaker:
also, updated to 2.1.0 SDK. this caused a bunch of errors in the provider to become apparent, so we cleaned those up. numerous fields listed in the schema were never actually set, and have been removed.


lgfa29 (approved):
:shipit: 

I update the SDK version to 2.4.0 since this PR has been waiting for review for a while. I also updated Nomad to 1.0.1.

I also ran a quick test to make sure that the changes to the `nomad_job` resource wouldn't impact existing jobs. I used the [Nomad E2E test suite](https://github.com/hashicorp/nomad/tree/master/e2e) but deployed the jobs using TF with the current released version of the provider (1.4.11).

I then updated my config to use a custom build from this PR and ran `terraform apply` again. There were no significant differences between in the [state diff](https://gist.github.com/lgfa29/f9673fee9a8e43111f47a0319f59be8c). 

One group disappeared, but I think that was part of the test, and one volume switched from `read_only = true` to `read_only = false`, but I think that's because my environment didn't have the specific CSI plugin properly setup. Other volumes kept their `read_only` state.

lgfa29 (changes requested):
Blocking this one for now because it modifies the schema of `mount_options` in `nomad_volume` in a way that will conflict with #188

lgfa29 (dismissed):

lgfa29 (approved):

PR #168, opened 2020-11-02
datasource: add Nomad scheduler config data source.

jrasell:
There are no tests, because they unfortunately clash with the scheduler config resource, which updates the config values meaning the datasource tests have a race condition output with that. I'll think how to better approach this, but it highlights why this resource should be used carefully.

The dependancy update is unrelated to this change, but is needed as a result of 438a0b8 so is included.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -13,6 +13,7 @@ exclude (
 
 require (
 	github.com/google/go-cmp v0.5.0
+	github.com/hashicorp/go-multierror v1.1.0
 	github.com/hashicorp/go-version v1.2.1
 	github.com/hashicorp/nomad v0.12.5-0.20201029140339-d6255129a300
 	github.com/hashicorp/nomad/api v0.0.0-20201028165800-38e23b62a770
--- a/nomad/data_source_scheduler_config.go
+++ b/nomad/data_source_scheduler_config.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/hashicorp/terraform-plugin-sdk/helper/resource"
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+	"github.com/hashicorp/terraform-provider-nomad/nomad/helper"
 )
 
 func dataSourceSchedulerConfig() *schema.Resource {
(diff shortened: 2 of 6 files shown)


website/docs/d/scheduler_config.html.md
```diff
@@ -0,0 +1,24 @@
...
+```hcl
+data "nomad_scheduler_config" "global" {}
+```
+
+## Attribute Reference
+
+The following attributes are exported:
+
+* `scheduler_algorithm` `(string)` - Specifies whether scheduler binpacks or spreads allocations on available nodes.
+* `preemption_config` `(map[string]bool)` - TOptions to enable preemption for various schedulers.
```
cgbaker, line 24:
```suggestion
* `preemption_config` `(map[string]bool)` - Options to enable preemption for various schedulers.
```

website/docs/d/scheduler_config.html.md
```diff
@@ -0,0 +1,24 @@
...
+layout: "nomad"
+page_title: "Nomad: nomad_scheduler_config"
+sidebar_current: "docs-nomad-datasource-scheduler-config"
+description: |-
+  Retrieve the cluster's scheduler configuration.
+---
+
+# nomad_scheduler_config
+
+Retrieve the cluster's scheduler configuration.
```
cgbaker, line 11:
```suggestion
Retrieve the cluster's [scheduler configuration](https://www.nomadproject.io/api-docs/operator#sample-response-3).
```

nomad/data_source_scheduler_config.go
```diff
@@ -0,0 +1,54 @@
...
+}
+
+func dataSourceSchedulerConfigRead(d *schema.ResourceData, meta interface{}) error {
+
+	client := meta.(ProviderConfig).client
+
+	schedCfg, _, err := client.Operator().SchedulerGetConfiguration(nil)
+	if err != nil {
+		return fmt.Errorf("failed to query scheduler config: %v", err)
+	}
```
cgbaker, line 37: @lgfa29 added a StateWriter utility on Friday that helps with these error checks: 
https://github.com/hashicorp/terraform-provider-nomad/pull/162/files#diff-3bd71acc5dfea73f22ad03c6d60d4f85ab09c8a1d818a1c4c148cfac12c5eab0R71-R86

go.mod
```diff
@@ -13,6 +13,7 @@ exclude (
 
 require (
 	github.com/google/go-cmp v0.5.0
+	github.com/hashicorp/go-multierror v1.1.0
```
cgbaker, line 16: go mod tidy?

cgbaker (approved):
cool.

the standard trick that we've used for testing these is a resource to set, then a dependent datasource should see the new value. e.g.,
https://github.com/hashicorp/terraform-provider-nomad/blob/master/nomad/data_source_acl_policy_test.go#L36-L48

did that not work here?

go.mod
jrasell: Negative; I did indeed try this. `go mod why` details it's because of the helper package:
```
# github.com/hashicorp/go-multierror
github.com/hashicorp/terraform-provider-nomad/nomad/helper
github.com/hashicorp/go-multierror
```

lgfa29: I probably forgot to commit this when I created the StateWriter helper..

lgfa29 (changes requested):
LGTM.

It's just missing an entry for the side menu in `website/nomad.erb` (we need a PR checklist...)

lgfa29 (approved):

PR #169, opened 2020-11-06
Fix issue where volume register would not set namespace

lgfa29:
When registering a volume in a specific namespace, the provider would fail with the error:

```
Error: Provider produced inconsistent result after apply

When applying changes to nomad_volume.ebs, provider
"registry.terraform.io/hashicorp/nomad" produced an unexpected new value: Root
resource was present, but now absent.

This is a bug in the provider, which should be reported in the provider's own
issue tracker.
```

This happened because the registration action was not setting the namespace in the request, but the read (and delete) do, so the volume would 404 on read and the state would be inconsistent.


cgbaker (approved):
LGTM

PR #174, opened 2020-12-01
added discussion for providing vault tokens to nomad job resources

cgbaker:


website/docs/index.html.markdown
```diff
@@ -91,3 +92,75 @@ resource "nomad_job" "nomad_eu" {
...
+* the token can be set in the `VAULT_TOKEN` environment variable when running `terraform apply`
+
+There are two problems that arise. The first is that we likely want to avoid putting
+Vault tokens into Terraform files where they may be inadvertently distributed. The second
+is that Nomad jobs might require different Vault tokens, each with access to a
+different set of policies. In this case, there are a few different strategies for
+managing the tokens and ensuring that the correct token is used for a particular
+job.
+
+One approach is to provider aliases, creating a Nomad Provider for each Vault token:
```
cgbaker, line 118:
```suggestion
One approach is to use provider aliases, creating a Nomad Provider for each Vault token:
```

website/docs/index.html.markdown
```diff
@@ -91,3 +92,75 @@ resource "nomad_job" "nomad_eu" {
...
+```
+
+The tokens can be passed from the command lines as variables:
+```bash
+$ terraform apply  -var vault_token_a=s.lraLq3axH9mkbdVRkWS6H06Q  \
+                   -var vault_token_b=s.koqvVqdAkG8yt7irxDdmIQiC
+```
+
+The downside here is that it requires creating multiple Nomad provider aliases
+and specifying the desired alias for every job resource. Another approach is inject
```
cgbaker, line 148:
```suggestion
and specifying the desired alias for every job resource. Another approach is to inject
```

website/docs/index.html.markdown
```diff
@@ -91,3 +92,75 @@ resource "nomad_job" "nomad_eu" {
...
+used in the Nomad job. When running the Nomad CLI, this token can be provided in
+a number of ways:
+* in the job spec using the [`vault_token`](https://www.nomadproject.io/docs/job-specification/job#vault_token) parameter
+* using the [`-vault-token`](https://www.nomadproject.io/docs/commands/job/run#vault-token) command line flag
+* setting the `VAULT_TOKEN` environment variable.
+
+When using the Nomad Provider to register Nomad jobs, the options are similar:
+* the token can be placed in the job spec in the [`nomad_job`](./resources/job) resource
+* the token can be [configured](#vault_token) on the Nomad Provider.
+* the token can be set in the `VAULT_TOKEN` environment variable when running `terraform apply`
```
angrycub, line 109: Minus is preferred.

```suggestion
- in the job spec using the [`vault_token`](https://www.nomadproject.io/docs/job-specification/job#vault_token) parameter
- using the [`-vault-token`](https://www.nomadproject.io/docs/commands/job/run#vault-token) command line flag
- setting the `VAULT_TOKEN` environment variable.

When using the Nomad Provider to register Nomad jobs, the options are similar:
- the token can be placed in the job spec in the [`nomad_job`](./resources/job) resource
- the token can be [configured](#vault_token) on the Nomad Provider.
- the token can be set in the `VAULT_TOKEN` environment variable when running `terraform apply`
```

angrycub (commented):
Nice content!

angrycub (dismissed):

PR #184, opened 2021-01-12
provider: support TLS certs configured by strings.

### LINKED ISSUE #181: Support for TLS Certificates as Strings
Currently, the Consul provider supports being given certificates for Mutual TLS inline (in a variable) or as a file path, in its `cert_pem` and `key_pem` fields. The Nomad provider lacks this functionality, as certificates can only be provided as paths. Matching this functionality would be very useful for environments like Terraform Cloud, where the only way to put a file onto a system would be to include it in version control.

This functionality was implemented in hashicorp/terraform-provider-consul#220, and whilst I plan to look at copying these changes to a PR on this repository, my Go and Terraform Provider Development experience is very limited, so that may not be a time efficient solution. Notably, the structure of unit tests seems to be a bit different on that repository, so this will need to be examined when migrating these changes. If the plan is to implement the same or very similar code to the aforementioned PR, I think it would be appropriate for @lawliet89 to give their approval, as it is ultimately their work.

As I say, I plan to look at making these changes myself, but I am happy to test someone else's implementation if they develop one.

jrasell:
This change adds a number of new provider config params that allow users to configure TLS certificates via strings, alongside the pre-existing file location method.

The provider schema ConflictsWith param is used to notify users when a configuration attempts to use two keys which affect the same TLS config param. In order to successfully use this, the existing DefaultFunc default values have been moved from an empty string to nil. This is because Terraform treats an empty string var as being populated, and therefore raises a conflict error whenever a new param is set. I tested this change locally to ensure it didn't cause any compatibility issues, and didn't find any. It is reassures that the Consul provider went through the same change.

The provider doesn't have much testing around its configuration, so I will open an issue to improve this in the future. In the meantime when testing locally I used the [Nomad dev tls cluster](https://github.com/hashicorp/nomad/tree/master/dev/tls_cluster) along with the Terraform code below to test. The local plugin is located within `~/.terraform.d/plugins/hashicorp.com/hashicorp/nomad/0.1/darwin_amd64/`.

```hcl
provider "nomad" {
  address  = "https://127.0.0.1:4646"
  ca_pem   = file("~/go/src/github.com/hashicorp/nomad/dev/tls_cluster/certs/nomad-ca.pem")
  cert_pem = file("~/go/src/github.com/hashicorp/nomad/dev/tls_cluster/certs/server.pem")
  key_pem  = file("~/go/src/github.com/hashicorp/nomad/dev/tls_cluster/certs/server-key.pem")
}

resource "nomad_namespace" "test" {
  name = "test"
}

terraform {
  required_providers {
    nomad = {
      versions = ["0.1"]
      source   = "hashicorp.com/hashicorp/nomad"
    }
  }
}

```
closes #181

### DIFF
--- a/nomad/provider.go
+++ b/nomad/provider.go
@@ -23,7 +23,6 @@ func Provider() terraform.ResourceProvider {
 				DefaultFunc: schema.EnvDefaultFunc("NOMAD_ADDR", nil),
 				Description: "URL of the root of the target Nomad agent.",
 			},
-
 			"region": {
 				Type:        schema.TypeString,
 				Optional:    true,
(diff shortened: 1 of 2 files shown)


lgfa29 (approved):
LGTM 👍

Issue #179, opened 2020-12-25
missing support for hcl2 jobspec


xinau:
### Terraform Version
```
$ terraform -v                 
Terraform v0.14.3
+ provider registry.terraform.io/hashicorp/nomad v1.4.11
```

### Nomad Version
```
$ nomad server members
Name         Address      Port  Status  Leader  Protocol  Build  Datacenter  Region
test.global  192.168.1.2  4648  alive   true    2         1.0.1  dc1         global
```

### Provider Configuration

```hcl
provider "nomad" {
  address   = "https://test.example.com:4646"
  ca_file   = "${path.root}/../private/nomad-ca.pem"
  cert_file = "${path.root}/../private/nomad-cli.pem"
  key_file  = "${path.root}/../private/nomad-cli-key.pem"
}
```

### Affected Resource(s)
- nomad_job
- nomad_job_parser

### Terraform Configuration Files
```hcl
resource "nomad_job" "main" {
  jobspec = file("${path.root}/jobspec.hcl")
}
```

```hcl
...
      template {
        data          = file("./prometheus.yml")
        destination   = "local/config/prometheus.yml"
        change_mode   = "signal"
        change_signal = "SIGHUP"
      }
...
```

### Expected Behavior

1 resource created.

### Actual Behavior

```console
$ terraform apply                                             

Error: error parsing job: &errors.errorString{s:"Unexpected response code: 400 (input.hcl:30,25-30: Error in function call; Call to function \"file\" failed: filesystem function disabled.\ninput.hcl:30,25-30: Unsuitable value type; Unsuitable value: value must be known\ninput.hcl:37,23-28: Error in function call; Call to function \"file\" failed: filesystem function disabled.\ninput.hcl:37,23-28: Unsuitable value type; Unsuitable value: value must be known\ninput.hcl:42,23-28: Error in function call; Call to function \"file\" failed: filesystem function disabled.\ninput.hcl:42,23-28: Unsuitable value type; Unsuitable value: value must be known\ninput.hcl:47,23-28: Error in function call; Call to function \"file\" failed: filesystem function disabled.\ninput.hcl:47,23-28: Unsuitable value type; Unsuitable value: value must be known)"}
```

### Steps to Reproduce
1. `terraform apply`

### Important Factoids

HCL2 Language Support

spol-ihq:
Same here,

Adding `variables` block at the top of the `nomad.hcl.template` and using `dynamic group {}`:

```
variables {
[...]

PR #185, opened 2021-01-12
Support HCL2 on nomad Go package

thiagonache:
Fix https://github.com/hashicorp/terraform-provider-nomad/issues/179

### DIFF
--- a/nomad/resource_job.go
+++ b/nomad/resource_job.go
@@ -11,6 +11,7 @@ import (
 	"time"
 
 	"github.com/hashicorp/nomad/api"
+	"github.com/hashicorp/nomad/jobspec"
 	"github.com/hashicorp/nomad/jobspec2"
 	"github.com/hashicorp/terraform-plugin-sdk/helper/resource"
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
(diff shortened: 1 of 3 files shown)


lgfa29:
Thanks so much for the PR @thiagonache!

It's good to see the tests passing now 🙂

Since the HCL2 jobspec parser is still not 100% compatible with the previous one, I think it would be better to have a config flag that could be set to indicate that the HCL2 parser should be used. 

Something like this:

```hcl
resource "nomad_job" "example" {
  jobspec = "..."
  hcl2 = true
}
```

This config would default to `false` to maintain backwards compatibility with existing Terraform files.

Let me know of you have any questions.

thiagonache:
@lgfa29 Please, let me know if that's okay.

lgfa29 (approved):
LGTM 👍 

thiagonache:
just curious, what's the release cadence of the provider?

lgfa29:
> just curious, what's the release cadence of the provider?

We don't have any specific cadence, it's usually driven by demand or if there is any critical bug fix.

Are you waiting on this feature? There are a few other PRs that should be merged soon, so we will probably cut a release in the coming days.

thiagonache:
That’s good enough. I can use my own build for couple days.

🤩

lgfa29:
FYI @thiagonache,

Turns out that loading local files in HCL2 is a bit tricky, so we changed how the `hcl2` parameter is specified in #192. 

Since your PR has not been released yet, this change won't be listed as a breaking change in our CHANGELOG, but if you are using a custom build you will need to update your Terraform config files.

thiagonache:
thx for letting me know!!!

PR #188, opened 2021-01-15
Fix `mount_options` in `nomad_volume` 

### LINKED ISSUE #182: Nomad Volume mount_options Issue
### Terraform Version
```
Terraform v0.14.3
+ provider registry.terraform.io/hashicorp/nomad v1.4.11
```

### Nomad Version
```
Name                 Address     Port  Status  Leader  Protocol  Build  Datacenter  Region
hashistack-1.global  10.0.2.127  4648  alive   false   2         1.0.1  dc1         global
hashistack-2.global  10.0.2.107  4648  alive   true    2         1.0.1  dc1         global
hashistack-3.global  10.0.2.179  4648  alive   false   2         1.0.1  dc1         global
```

### Provider Configuration

```hcl
provider "nomad" {
  address = "http://10.0.2.127:4646"
  region  = "global"
}
```

### Environment Variables
N/A

### Affected Resource(s)
- nomad_volume

If this issue appears to affect multiple resources, it may be an issue with Terraform's core, so please mention this.

```hcl
resource "nomad_volume" "registry_volume" {
  type                  = "csi"
  plugin_id             = "truenas"
  volume_id             = "registry"
  name                  = "registry"
  external_id           = "registry"
  access_mode           = "single-node-writer"
  attachment_mode       = "file-system"
  deregister_on_destroy = true

  mount_options = {
    fs_type = "nfs"
  }

  context = {
    node_attach_driver = "nfs"
    provisioner_driver = "freenas-nfs"
    server             = "freenas.zt.domain.tld"
    share              = "/mnt/data/nomad/vols/registry"
  }
}

```

### Debug Output
https://gist.github.com/aditsachde/930e1b8121e5819aec25e832243c3f0d

### Expected Behavior
When the unchanged volume is applied again, Terraform should not update anything.

### Actual Behavior
When the unchanged volume is applied again, Terraform will try to update the mount options. This is an issue as Nomad will reject the update if the volume is already in use.

### Steps to Reproduce
1. `terraform apply`
2. `terraform apply` again

### Important Factoids
N/A

### References
Are there any other GitHub issues (open or closed) or Pull Requests that should be linked here? For example:
- GH-145

lgfa29:
The `nomad_volume` resource had `mount_options` as a map with complex types, which Terraform can't process properly. The Nomad API also redacts `mount_options` from its response since it may contain sensitive values.

This PR changes the field type to a list with only 1 item to allow passing the sub-fields from the config file. It also doesn't try to read back `mount_options` from the API since it will always be `nil`.

Since the schema was modified, a migration function was necessary. 

Fixes #182

### DIFF
--- a/nomad/resource_volume.go
+++ b/nomad/resource_volume.go
@@ -104,16 +104,20 @@ func resourceVolume() *schema.Resource {
 			"mount_options": {
 				Description: "Options for mounting 'block-device' volumes without a pre-formatted file system.",
 				Optional:    true,
-				Type:        schema.TypeMap,
+				Type:        schema.TypeList,
+				MaxItems:    1,
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
 						"fs_type": {
 							Description: "The file system type.",
 							Type:        schema.TypeString,
+							Optional:    true,
 						},
 						"mount_flags": {
 							Description: "The flags passed to mount.",
 							Type:        schema.TypeList,
+							Elem:        &schema.Schema{Type: schema.TypeString},
+							Optional:    true,
 						},
 					},
 				},
@@ -194,6 +198,14 @@ func resourceVolume() *schema.Resource {
 				Type:     schema.TypeBool,
 			},
 		},
+		SchemaVersion: 1,
+		StateUpgraders: []schema.StateUpgrader{
+			{
+				Type:    resourceVolumeResourceV0().CoreConfigSchema().ImpliedType(),
+				Upgrade: resourceVolumeStateUpgradeV0,
+				Version: 0,
+			},
+		},
 	}
 }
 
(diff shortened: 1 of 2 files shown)


nomad/resource_volume.go
```diff
@@ -322,7 +339,192 @@ func resourceVolumeRead(d *schema.ResourceData, meta interface{}) error {
 	d.Set("nodes_healthy", volume.NodesHealthy)
 	d.Set("nodes_expected", volume.NodesExpected)
 	d.Set("schedulable", volume.Schedulable)
-	d.Set("mount_options", volume.MountOptions)
+	// The Nomad API redacts `mount_options` and `secrets`, so we can't set
+	// them here.
```
cgbaker, line 343:
```suggestion
	// The Nomad API redacts `mount_options` and `secrets`, so we don't update them
	// with the response payload; they will remain as is.
```

cgbaker (approved):
good job, especially on the upgrade.

lgfa29:
@cgbaker can I get a ✅?  😄

cgbaker:
that's so weird, i thought i did.... has github changed the order of the combo box or somethign?

cgbaker (approved):
approved!!!

PR #189, opened 2021-01-15
Add support for HTTP basic auth

### LINKED ISSUE #180: HTTP Basic Auth Support
It would be really useful if the provider supported HTTP Basic Auth in the same way [hashicorp/terraform-provider-consul](https://github.com/hashicorp/terraform-provider-consul/) does. Whilst ACLs should provide the same level of security, an additional layer would be useful for workloads such as using the provider in Terraform Cloud to manage on-prem Nomad deployments. I'm sadly not familiar enough with Terraform provider development to implement this myself, but I'm happy to test an implementation from someone else if one is made.

Edit: As the module supports mutual TLS (or at least appears to from the docs), this is not the only solution, however basic auth can be a way of achieving a slightly lower level of security but a lot faster for people bootstrapping their first environment.

lgfa29:
This PR adds the `http_auth` provider config key to allow passing HTTP basic auth settings to the Nomad client.

Closes #180


jrasell (approved):
LGTM

PR #197, opened 2021-01-21
ci: update hashistack deps

lgfa29:
The HCL2 tests target Nomad 1.0.0+, so they weren't running on CI.

### DIFF
--- a/scripts/getconsul.sh
+++ b/scripts/getconsul.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-CONSUL_BINARY=https://releases.hashicorp.com/consul/1.8.5+ent/consul_1.8.5+ent_linux_amd64.zip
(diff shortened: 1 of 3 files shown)


jrasell (approved):
LGTM

It's worth noting we could grab the latest versions from checkpoint; although I don't know if we would want this.

lgfa29:
> LGTM
> 
> It's worth noting we could grab the latest versions from checkpoint; although I don't know if we would want this.

Yeah, I thought about that, but left it pinned for now since this runs in every commit, and we probably don't want it to break for external factors. 

I think the best approach would be to have a more comprehensive test matrix that runs nightly (or on merges to `master`?) across multiple version combinations, but my [WIP](https://github.com/hashicorp/terraform-provider-nomad/tree/tmp-testacc-docker/scripts/tftest) is not quite there yet.

PR #203, opened 2021-03-09
Add headers to provider

Mongey:


Mongey:
I think I probably meant to add tests/docs, and then forgot to follow back up 😅

cgbaker:
@Mongey , i don't mind picking this up and adding tests/docs.

cgbaker:
thanks for the pr, @Mongey ; i've added some tests and documentation.

jrasell (dismissed):
LGTM although I am not sure what is causing the CI failure. I believe `go mod tidy` could be run to clean out some entries in the sum file, but this isn't critical.

cgbaker:
i broke something in `main.go` when i change one of the returns, fixing it now

PR #211, opened 2021-03-31
Allow passing HCL2 variables to job

### LINKED ISSUE #195: Allow Passing Variables Input With hcl2 `nomad_job`
### Terraform Version
`Terraform v0.14.5`

### Nomad Version
`1.0.1+ent`

### Provider Configuration

```hcl
provider "nomad" {
  address = data.terraform_remote_state.infrastructure.outputs.nomad_address
}

```

### Environment Variables
Do you have any Nomad specific environment variable set in the machine running Terraform?

N/A

### Affected Resource(s)
- nomad_job

If this issue appears to affect multiple resources, it may be an issue with Terraform's core, so please mention this.

```hcl
resource "nomad_job" "test" {
  jobspec = file("${path.module}/job.nomad")
  hcl2 {
    enabled = true
    vars = {
      test = "test1"
    }
  }
}
```

And the contrived `job.nomad`:

```hcl
variable "test" {
  type    = string
  default = "test fail"
}

job "docs" {

  datacenters = [
    "us-west-2",
  ]

  group "example" {
    task "server" {
      driver = "docker"

      config {
        image = "hashicorp/http-echo"

        args = [
          "-listen",
          ":5678",
          "-text",
          "hello world ${var.test}",
        ]
      }

      resources {
        network {
          mbits = 10

          port "http" {
            static = "5678"
          }
        }
      }
    }
  }
}

```

### Expected Behavior
It would be nice to be able pass in a map of variables with Nomad hcl2 jobspecs. For context, we end up with a lot of `jobspec = templatefile(...)` `nomad_job` definitions and it would be nice to use a more nomad-aware method of passing in variables from terraform. The `jobspec = templatefile(...)` arrangement means we end up with partial jobspecs checked into VCS that can't be readily used outside the context of terraform runs (excepting kludgey things like including the templatefile-rendered jobspec as a TF output or some such).

I sketched out a quick implementation that appears to technically work / pass in variables as desired here: https://github.com/hashicorp/terraform-provider-nomad/compare/master...jeffwecan:hcl2_vars?expand=1

### Actual Behavior
What actually happened?

```shellsession
%▶ tf apply

Error: Unsupported argument

  on main.tf line 10, in resource "nomad_job" "test":
  10:     vars = {

An argument named "vars" is not expected here
```

lgfa29:
Add new attribute to the `hcl2` field in the `nomad_job` resource to allow passing extra HCL2 variable values.

Docs preview:
image

Closes #195

### DIFF
--- a/nomad/resource_job.go
+++ b/nomad/resource_job.go
@@ -715,7 +715,7 @@ func parseHCL2JobParserConfig(raw interface{}) (HCL2JobParserConfig, error) {
 	if vars, ok := hcl2Map["vars"].(map[string]interface{}); ok {
 		config.Vars = make(map[string]string)
 		for k, v := range vars {
-			config.Vars[k] = fmt.Sprintf("%s", v)
+			config.Vars[k] = v.(string)
 		}
 	}
 
--- a/nomad/resource_job.go
+++ b/nomad/resource_job.go
@@ -92,6 +92,11 @@ func resourceJob() *schema.Resource {
 							Optional:    true,
 							Default:     false,
 						},
+						"vars": {
+							Description: "Additional variables to use when templating the job with HCL2",
+							Type:        schema.TypeMap,
+							Optional:    true,
+						},
 					},
 				},
 			},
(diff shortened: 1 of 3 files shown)


cgbaker (approved):
👍

Issue #222, opened 2021-05-10
Apply / destroy with Terraform 0.15 crash with "concurrent map read and map write"


multani:
### Terraform Version

```
$ terraform0.15 -v
Terraform v0.15.3
on linux_amd64
+ provider registry.terraform.io/hashicorp/nomad v1.4.14
```

### Nomad Version

```
$ nomad server members                                     
Name            Address    Port  Status  Leader  Protocol  Build  Datacenter  Region
alumine.global  127.0.0.1  4648  alive   true    2         1.0.2  dc1         global
```

it also happens with an older version of Nomad. (I don't think the Nomad version matters in this issue)

### Provider Configuration
Which values are you setting in the provider configuration? 

```hcl
provider "nomad" {                                                                                                                                                                                                                                                                                                           
  address   = "http://localhost:4646"
}
```

### Environment Variables

* `NOMAD_TOKEN`:  set after running `nomad acl bootstrap`.

### Affected Resource(s)

My reproducible use-case works with `nomad_acl_token`, but I think it affects other resources as well:

- `nomad_acl_token`

### Terraform Configuration Files

```hcl
provider "nomad" {
  address = "http://localhost:4646"
}

resource "nomad_acl_token" "tok" {
  count = 10
  name  = "tok tok"
  type  = "management"
}
```

I used `count = 10`, but I think any configuration with concurrent updates on a single (perhaps several?) resources should exhibit the same issue.

### Debug Output

See [`tf.log`](https://gist.github.com/multani/00078402d51ab604adc6e7d8e67f06cb#file-tf-log)

### Panic Output

See [`output.log`](https://gist.github.com/multani/00078402d51ab604adc6e7d8e67f06cb#file-output-log)

### Expected Behavior

It should not crash.

### Actual Behavior

It crashes with a `fatal error: concurrent map read and map write` error.

It doesn't crash using Terraform 0.14.

### Steps to Reproduce

How to reproduce:

* run `nomad agent -dev -acl-enabled`
* run `nomad acl bootstrap`
* run `export NOMAD_TOKEN=XXX`, where XXX is the Secret ID from previous step
* run in a loop: `terraform apply` followed by `terraform destroy`, eventually it should crash. If not, try to increase the `count` of the token resource.

### Important Factoids

* It also crashes during the "destroy" operation.
* It used to work well with Terraform 0.14
[...]

Issue #221, opened 2021-05-08
fatal error: concurrent map read and map write


elprans:
### Terraform Version

```
Terraform v0.15.1
on linux_amd64
+ provider registry.terraform.io/hashicorp/local v2.1.0
+ provider registry.terraform.io/hashicorp/nomad v1.4.14
+ provider registry.terraform.io/hashicorp/null v3.1.0
+ provider registry.terraform.io/hashicorp/template v2.2.0
+ provider registry.terraform.local/dmacvicar/libvirt v0.6.3
```

### Nomad Version

```
Name           Address      Port  Status  Leader  Protocol  Build        Datacenter  Region
nomad0.global  10.17.3.73   4648  alive   true    2         1.1.0-beta1  local-1     global
nomad1.global  10.17.3.104  4648  alive   false   2         1.1.0-beta1  local-1     global
nomad2.global  10.17.3.238  4648  alive   false   2         1.1.0-beta1  local-1     global
```

### Provider Configuration

```hcl
provider "nomad" {
  address = "http://nomad.service.${local.datacenter}.consul:4646"
}
```

### Environment Variables

```sh
$ env | grep NOMAD_
NOMAD_ADDR=http://nomad.service.local-1.consul:4646
```

### Affected Resource(s)

- nomad_job

### Panic Output

<details><summary>PANIC OUTPUT</summary>
<p>

```
module.nomad-ebs.data.nomad_plugin.ebs: Reading...
╷
│ Error: Plugin did not respond
│ 
│   with nomad_job.loki,
│   on main.tf line 99, in resource "nomad_job" "loki":
│   99: resource "nomad_job" "loki" {
│ 
│ The plugin encountered an error, and failed to respond to the plugin.(*GRPCProvider).PlanResourceChange call. The plugin logs may contain more details.
╵
╷
│ Error: Plugin did not respond
│ 
│   with nomad_job.grafana,
│   on main.tf line 114, in resource "nomad_job" "grafana":
│  114: resource "nomad_job" "grafana" {
│ 
│ The plugin encountered an error, and failed to respond to the plugin.(*GRPCProvider).PlanResourceChange call. The plugin logs may contain more details.
╵
╷
│ Error: Plugin did not respond
│ 
│   with module.nomad-ebs.nomad_job.ebs_controller,
│   on ../modules/nomad-ebs/nomad-ebs.tf line 18, in resource "nomad_job" "ebs_controller":
[...]

Issue #220, opened 2021-05-04
Support for job.group.consul block in nomad_job


optiz0r:
### Terraform Version

Terraform: 0.14.11
Terraform-provider-nomad: 1.4.14

### Nomad Version

1.1.0-beta1

### Provider Configuration

```hcl
provider "nomad" {}
```

### Environment Variables

`NOMAD_ADDR` and `NOMAD_TOKEN` are set appropriately

### Affected Resource(s)

- nomad_job

### Terraform Configuration Files
```hcl
resource "nomad_job" "test" {
  jobspec = <<-EOT
   ##..
  group "test" {
    consul {
      namespace = "default"
    }
  }
  # ..
  EOT
  hcl2 {
    enabled = true
  }
}
```

### Expected Behavior

Job should be parsed correctly and submitted to Nomad for execution

### Actual Behavior

```
Terraform v0.14.11
Configuring remote state backend...
Initializing Terraform configuration...
# ..

Error: error parsing jobspec: :74,5-11: Unsupported block type; Blocks of type "consul" are not expected here.

  on nomad_consul_test.tf line 1, in resource "nomad_job" "test":
   1: resource "nomad_job" "test" {
```

### Steps to Reproduce

1. `terraform plan`

### Important Factoids

Likely the provider needs rebuilding against the nomad 1.1.0-beta+ SDK to support the new specification jobs in 1.1.0

jrasell:
Thanks @optiz0r, this is something we will certainly get updated and sorted.

Issue #219, opened 2021-05-04
resource: add MemoryOversubscriptionEnabled param to SchedulerConfig


jrasell:
The new memory oversubscription feature introduces a new [API parameter](https://www.nomadproject.io/api-docs/operator/scheduler#memoryoversubscriptionenabled) which should be made available within the Terraform provider.

PR #226, opened 2021-05-11
Update Nomad dependency

### LINKED ISSUE #214: gateway -> invalid key: terminating
### Terraform Version
This is being built in CI/CD using a docker container (hashicorp/terraform:0.14.9)

### Nomad Version
Running the `nomad server members` command shows all servers are using Build 1.0.4

### Provider Configuration

```hcl
provider "nomad" {}
```

Provider-wise, the logs indicate that the provider is using the latest one which supports the new 1.0.4 syntax.  I've gone and pinned the version to this latest one just in case.
```
Initializing provider plugins...
- Finding hashicorp/nomad versions matching "1.4.14"...
- Installing hashicorp/nomad v1.4.14...
- Installed hashicorp/nomad v1.4.14 (signed by HashiCorp)
```
### Environment Variables
Do you have any Nomad specific environment variable set in the machine running Terraform?

```sh
NOMAD_ADDR
NOMAD_REGION
NOMAD_TOKEN
```

### Affected Resource(s)
- nomad_job

```hcl
job "test" {
  region      = "testregion"
  datacenters = ["us-west-2"]
  type        = "service"

  vault {
    policies = ["nomad_vault"]
  }

  constraint {
    attribute = "$${attr.kernel.name}"
    value     = "linux"
  }

  group "gateway" {
    network {
      mode = "bridge"
    }

    service {
      name = "terminating-gateway"
      connect {
        gateway {
          proxy {

          }
          terminating {
            service {
              name = "tgw-service"
            }
          } 
        }
      }
    }
  } 
  
  ...other services...
}
```

### Expected Behavior
The Terraform Nomad job should run a plan without issue.

### Actual Behavior
The job does not understand the 1.0.4 Terminating Gateway syntax.

```
Error: error parsing jobspec: error parsing 'job': group: 'gateway', service (0): 'terminating-gateway', gateway, 1 error occurred:
	* gateway -> invalid key: terminating
```

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. `terraform plan`

### Important Factoids
ACL is enabled and the token is being provided to Nomad via the `NOMAD_TOKEN` env var.  I have gone in and created a terminating gateway by manually altering the JSON of the job, so I can confirm that terminating gateways can be created using my Nomad cluster - it is just the Nomad provider that does not understand it.  Also, if I just comment out the gateway group the plan completes without issue, so there is no issue communicating with Nomad itself.

lgfa29:
Update Nomad dependency to add support for new jobspec fields.

It also fixes an upstream problem introduced when #203 added support for additional headers for the provider, where a bug in the Nomad `api` package that caused a panic whenever concurrent access to the header map happened.

A new test and a broken commit are provided to illustrate that this fixes the problem.

Closes #214 #215 #219 #220 #221 #222


nomad/data_source_scheduler_config_test.go
```diff
@@ -13,28 +13,33 @@ func TestAccDataSourceSchedulerConfig_basic(t *testing.T) {
 		CheckDestroy: testFinalConfiguration,
 		Steps: []resource.TestStep{
 			{
-				Config: testAccNomadSchedulerConfigSpread,
+				Config: testAccNomadDataSourceSchedulerConfigS,
 				Check: resource.ComposeTestCheckFunc(
```
lgfa29, line 17: This test was broken and it seems to be flaky....I think it's clashing the the resource test since they are acting on the same cluster.

jrasell: Lets raise a follow up issue if there isn't one to track this.

jrasell (dismissed):
1.1.0 also adds `cores` into the resources stanza; it would be worth adding a test for this as a follow up also. LGTM though!

nomad/data_source_scheduler_config_test.go
lgfa29: There's this issue: https://github.com/hashicorp/terraform-provider-nomad/issues/163 and previously failed attempt to fix them https://github.com/hashicorp/terraform-provider-nomad/pull/173 😬

But now I'm thinking of combining the two into a since test case. They are basically doing the same thing: test resource to write, test data source to read.

lgfa29: Using a single test seems to have solved 🎉

jrasell (approved):
LGTM

PR #233, opened 2021-05-27
Migrate acceptance tests from Travis to Actions

mdeggies:
Travis-ci.org is being deprecated at the end of this week, so we're moving things to GitHub actions. Acceptance tests are run in actions using go 1.16.4 and tip on the ubuntu executor, and product versions were upgraded to latest.


lgfa29 (changes requested):
Thanks @mdeggies!

The `NOMAD_LICENSE` secret is not being loaded as an environment variable for some reason. I tried a few things, but no luck.

I added a step to upload the Nomad, Consul and Vault logs to an artifact on failure to make debugging easier.

One thing that I thought about doing was to pass the env var as a script input, so something like:

```yml
...
run: ./scripts/start-nomad.sh "$NOMAD_LICENSE"
...
```

```sh
# ./scripts/start-nomad.sh
...
export NOMAD_LICENSE=$1
...
```

But that could leak the secret?

mdeggies:
Hey @lgfa29, thanks for reviewing. I can investigate later today- I was able to get this working without needing to pass the license value to the bash script. I wonder if your permissions on this repo may be preventing your runs from using the github actions secret?

lgfa29:
Thanks! 

It was working before because it wasn't downloading an enterprise binary 😅 
https://github.com/hashicorp/terraform-provider-nomad/pull/233/commits/d689745f58238d75b8b93b0508be0969f8ed7926#diff-f49a6c12282815b6231708ca2710e485f9ba859903a4d3d29c184e9ca2fd662bL3

lgfa29 (approved):
Love that ✅!

Thanks for getting this migration done @mdeggies 🙂

PR #235, opened 2021-06-04
Wait for CSI plugin nodes to be healthy

lgfa29:
Following up on #234, this PR reads the number of expected nodes from the API result and also blocks until all nodes are healthy.


jrasell (approved):
LGTM nice change!

PR #238, opened 2021-06-16
Add nomad_job_v2 resource using terraform-plugin-go

remilapeyre:


meyerbro:
This is a very interesting and important work. This would mean that the TF PLAN could show only the differences but not a huge string with "+" and "-" in some lines... Any chance to get this PR reviewed? Thanks for this good work @remilapeyre !

lgfa29:
Despite @remilapeyre brave efforts, this approach turned out to be impossible to manage over time as it requires keeping the Terraform spec for jobs in sync with Nomad. 

We should investigate the advancements in the [Terraform plugin development ecosystem](https://developer.hashicorp.com/terraform/plugin/terraform-plugin-protocol) to come up with a better solution to #1.

Thank you again @remilapeyre for all the hard work on this!

lgfa29:
Hi @jorgemarey 👋 

Thanks for the input. We have received questions about a Nomad Pack + Terraform integration, but we haven't nailed down a design for that quite yet.

Would you mind moving this comment to a new issue so we can track it and discussed it more broadly? 

Thank you!

jorgemarey:
Created in https://github.com/hashicorp/terraform-provider-nomad/issues/409

PR #239, opened 2021-06-29
Don't require a Nomad license to run GitHub Actions

remilapeyre:
Also skip TestResourceJob_consulConnect() if Consul is not running.

### DIFF
--- a/nomad/provider_test.go
+++ b/nomad/provider_test.go
@@ -127,6 +127,21 @@ func testCheckVaultEnabled(t *testing.T) {
 	}
 }
 
+func testCheckConsulEnabled(t *testing.T) {
+	client := testProvider.Meta().(ProviderConfig).client
+	consulEnabled := false
+	if nodes, _, err := client.Nodes().List(nil); err == nil && len(nodes) > 0 {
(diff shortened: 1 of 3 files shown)


lgfa29:
Hi @remilapeyre thanks for the PR!

Unfortunately we need to load the ENT version of Nomad to make sure some of the tests run. The license was supposed to be loaded automatically since https://github.com/hashicorp/terraform-provider-nomad/pull/233. Maybe your fork doesn't have this PR?

lgfa29 (approved):
Oh, I think the problem is that secrets are not loaded on runners from forks

> Note: With the exception of GITHUB_TOKEN, secrets are not passed to the runner when a workflow is triggered from a forked repository.

https://docs.github.com/en/actions/reference/encrypted-secrets#using-encrypted-secrets-in-a-workflow

Then your changes make sense, since it will use the OSS version if the license env var is not found (which will happen on forks).

I will merge this and see if the enterprise tests run on `main`  😄 

Thanks again!

remilapeyre:
> Oh, I think the problem is that secrets are not loaded on runners from forks
> 
> > Note: With the exception of GITHUB_TOKEN, secrets are not passed to the runner when a workflow is triggered from a forked repository.

Yes, that's exactly that thanks!

PR #240, opened 2021-06-29
ci: fix nomad license env var check

lgfa29:
The previous check would fail with this error:

```
./scripts/getnomad.sh: line 6: [: missing `]'
./scripts/getnomad.sh: line 6: -n: command not found
```


jrasell (approved):
LGTM

PR #261, opened 2022-02-15
feat: add vault and consul token to job resource

anth0d:
There is an implicit cyclic dependency between providers when you try to use dynamic (i.e., generated by another provider) `vault_token` and `consul_token` in your Nomad provider configuration. This change introduces the ability to provide `vault_token` and `consul_token` as part of the `nomad_job` resource, so that dynamic token values can be provided by `vault` and `consul` providers respectively without hitting the [provider dependency issue](https://github.com/hashicorp/terraform/issues/2430).

Let me know how I can ensure the registry docs are also up to date with this change, and let me know if anything else is needed! Thanks!

### DIFF
--- a/nomad/resource_job.go
+++ b/nomad/resource_job.go
@@ -167,6 +167,20 @@ func resourceJob() *schema.Resource {
 				Optional:    true,
 				Type:        schema.TypeBool,
 			},
+
+			"consul_token": {
+				Description: "The Consul token used to submit this job.",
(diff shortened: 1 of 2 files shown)


lgfa29 (approved):
Thanks @anth0d, and sorry for the delay in getting this reviewed. 

I pushed a commit adding a note to the docs about storing these sensitive values in the state.

Issue #260, opened 2022-01-29
CSI volume created using nomad_external_volume resource seems to be missing mount_flags


CarbonCollins:
Hi there,

Thank you for opening an issue. Please note that we try to keep the Terraform issue tracker reserved for bug reports and feature requests. For general usage questions, please see: https://www.terraform.io/community.html.

### Terraform Version
```
Terraform v1.1.3
on linux_amd64
+ provider registry.terraform.io/hashicorp/nomad v1.4.15
+ provider registry.terraform.io/hashicorp/vault v2.24.1
```

### Nomad Version
```1.2.4```

### Provider Configuration
Which values are you setting in the provider configuration? 

```hcl
provider "nomad" {}
provider "vault" {}
```

### Environment Variables
Do you have any Nomad specific environment variable set in the machine running Terraform?

```sh
NOMAD_REGION=global
NOMAD_TOKEN=[redacted]
NOMAD_ADDR=https://[redacted]
```

### Affected Resource(s)
Please list the resources as a list, for example:
- nomad_external_volume

If this issue appears to affect multiple resources, it may be an issue with Terraform's core, so please mention this.

### Terraform Configuration Files
```hcl
locals {
  plugin_id = "axion-proxima"
  volume_id = "plex-data"
}

data "nomad_plugin" "axion_proxima" {
  plugin_id        = local.plugin_id
  wait_for_healthy = false
}

data "vault_generic_secret" "axion_proxima_auth" {
  path = var.axion_proxima_auth_secret_path
}

resource "nomad_external_volume" "plex_data" {
  depends_on  = [data.nomad_plugin.axion_proxima]

  type        = "csi"
  namespace   = "default"
  plugin_id   = local.plugin_id

  name        = local.volume_id
  volume_id   = local.volume_id

  capacity_min = "1 GiB"
  capacity_max = "20 GiB"

  capability {
    access_mode     = "single-node-reader-only"
    attachment_mode = "file-system"
  }

  capability {
    access_mode     = "single-node-writer"
    attachment_mode = "file-system"
  }

  mount_options {
    fs_type = "cifs"
    mount_flags = [
        "vers=3",
        format("uid=%s", var.user_id),
        format("gid=%s", var.group_id),
        "nolock",
        format("username=%s", data.vault_generic_secret.axion_proxima_auth.data["USER"]),
        format("password=%s", data.vault_generic_secret.axion_proxima_auth.data["PASS"])
    ]
  }
}

```

[...]

PR #266, opened 2022-04-25
resource/external_volume: fix mount flags parsing

tgross:
Fixes https://github.com/hashicorp/terraform-provider-nomad/issues/260

The mount flags parsing tries to cast directly to a slice of strings,
but this silently fails and leaves us with empty mount flags.

---

I've tested this manually but I need to brush up on how to write Terraform tests because I don't quite understand how it's working in this repo. Putting the PR up for review while I work on that.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ IMPROVEMENTS:
 * resource/nomad_job: allow passing Consul and Vault token through the job resource ([#261](https://github.com/hashicorp/terraform-provider-nomad/pull/261))
 
 BUG FIXES:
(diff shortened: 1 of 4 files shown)


lgfa29 (approved):
Thanks @tgross!

There's definitely a little bit of art in writing Terraform tests, and this resource didn't have any since it requires a cluster with CSI configured. 

I added some basic testing based on the `hostpath` demo since it's fairly easy to get it running locally with Vagrant.

Though I don't think they will be able to run in CI since it requires privileged Docker access. I will give it a try in a follow-up PR.

PR #271, opened 2022-06-08
provider: don't load namespace from env var

### LINKED ISSUE #267: nomad_job is put in a nonexistent namespace by default
### Terraform Version
Terraform v1.1.9
on linux_amd64
+ provider registry.terraform.io/hashicorp/nomad v1.4.16

### Nomad Version
Nomad v1.2.6 or v1.3.0

### Provider Configuration

```hcl
provider "nomad" {
  address = var.NOMAD_ADDR
  region  = "global"
}
```

### Affected Resource(s)

- nomad_job

```hcl
resource "nomad_job" "webapp" {
  jobspec = file("${path.module}/jobs/webapp.nomad")
  hcl2 {
    enabled = true
  }
}
```

### Debug Output
- TF_LOG_PROVIDER output: https://gist.github.com/dysonfrost/17afb60e0da0b56b926c465f19aee616
- Terraform plan output: https://gist.github.com/dysonfrost/663bc6458cf05f456d8844c130ba1166
- Nomad log sample output: https://gist.github.com/dysonfrost/3fa04875fb85f85bf8dd516d5f7d08d8

### Expected Behavior
Webapp job should have been put in the `default` namespace

### Actual Behavior
Terraform makes a request to put the job in the `isolation` namespace

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. `terraform apply`

### Important Factoids

Until last week I had no issue with this configuration, but since today I cannot create/modify a `nomad_job` resource.
The `isolation` namespace doesn't exist in the terraform state and nomad namespaces list.
Same result when I specifically declare another namespace that already exists.
I am using Terraform Cloud, also tried to run the job directly with Nomad 1.2.6 and 1.3.0, it works as expected.

lgfa29:
The Nomad provider doesn't have a global Namespace configuration as each
resource may target a different namespace, but the Nomad API client
default configuration would load a namespace value from the
`NOMAD_NAMESPACE` automatically.

This caused issues in environments where Terraform runs within a Nomad
allocation, where the value for `NOMAD_NAMESPACE` is the namespace of the
allocation.

Since the namespace value was never supposed to be set at the provider
level, this change makes sure the Nomad client namespace is always
empty.

Closes #267


jrasell (approved):
LGTM! Thanks for looking into this and finding such as nice solution.

PR #272, opened 2022-06-09
ci: stop Nomad before uploading logs

lgfa29:
I think the `actions/upload-artifact` may fail if the Nomad server is still writing to the log file. It's a big guess (I have no idea why [this](https://github.com/hashicorp/terraform-provider-nomad/runs/6814161240?check_suite_focus=true) failed), but it seems like the right order of operations regardless.


DerekStrickland (approved):
:shipit:

PR #280, opened 2022-08-25
provider: add load_namespace_env_var configuration

lgfa29:
Allow users to opt-in into loading the NOMAD_NAMESPACE environment
variable. #271 removed this, as this was not intentional, but this value
is used in some workflows.


jrasell (approved):
LGTM!

PR #282, opened 2022-08-29
deps: update Nomad to 1.3.4

lgfa29:


jrasell (commented):
I think we need to update the version reference in the go.mod file to accommodate Nomad's use of generics which will fix the failing test.

lgfa29:
I don't know what's happening with that log upload 500 errors 🤷

PR #284, opened 2022-09-05
acl: add ACL Role resource and data sources

jrasell:
This change adds a new `nomad_acl_role` resource along with `nomad_acl_role` and `nomad_acl_roles` datasources.

In order to raise the change, the Nomad dependency has been updated to latest. This should be fixed to a release tag once 1.4.0 is released.

The tests do not currently run because the CI process requires a released version of Nomad :D. We can either merge this as is, or hold off until 1.4.0; but raising the PR now for early review.


lgfa29 (approved):
:shipit:

PR #289, opened 2022-09-22
ci: update test to 1.4

lgfa29:


nomad/resource_job_test.go
```diff
@@ -1736,16 +1735,17 @@ func testResourceJob_csiControllerCheck(s *terraform.State) error {
 		return fmt.Errorf("task group %s not found", taskGroupName)
 	}
 
-	expCSIPluginConfig := api.TaskCSIPluginConfig{}
-	json.Unmarshal([]byte(`{
-        "ID": "aws-ebs0",
-        "Type": "controller",
-        "MountDir": "/csi"
-	}`), &expCSIPluginConfig)
```
lgfa29, line 1744: I'm not sure why we used to do this JSON decode thing, I found it a little harder to update tests since you can't rely on the actual struct information for things like autocomplete and code navigation.

scripts/getnomad.sh
```diff
@@ -2,7 +2,7 @@
 
 set -e
 
-NOMAD_VERSION=1.3.1
+NOMAD_VERSION=1.4.0-beta.1
```
lgfa29, line 5: We should update this after GA, but at least this PR gives a starting point for any work that needs to be done before that like #286 and #284.

jrasell (dismissed):

jrasell:
I pushed an update to move the CI install version to v1.4.1

jrasell (approved):

lgfa29:
> I pushed an update to move the CI install version to v1.4.1

Awesome, thanks!

PR #304, opened 2023-01-19
Added support for nomad variables

iluminae:
Here is a resource/datasource for interacting with nomad variables (v1.4+). I did not get to the tests yet, but I have tested this manually against v1.4.3.

### DIFF
--- a/nomad/resource_variable.go
+++ b/nomad/resource_variable.go
@@ -1,0 +1,101 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+package nomad
+
+import (
+	"fmt"
+	"log"
+
+	"github.com/hashicorp/nomad/api"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+)
+
+func resourceVariable() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceVariableWrite,
+		Update: resourceVariableWrite,
+		Delete: resourceVariableDelete,
+		Read:   resourceVariableRead,
+		Exists: resourceVariableExists,
+
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"path": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"namespace": {
+				Description: "Variable namespace filter",
+				Type:        schema.TypeString,
+				Optional:    true,
+				Default:     api.DefaultNamespace,
+			},
+			"items": {
+				Type:      schema.TypeMap,
+				Required:  true,
+				Sensitive: true,
+			},
+		},
+	}
+}
+
+func resourceVariableWrite(d *schema.ResourceData, meta any) error {
+	client := meta.(ProviderConfig).client
+
+	variable := api.Variable{
+		Namespace: d.Get("namespace").(string),
+		Path:      d.Get("path").(string),
+		Items:     make(map[string]string),
+	}
+	for name, value := range d.Get("items").(map[string]any) {
+		variable.Items[name] = value.(string)
+	}
+
+	log.Printf("[DEBUG] Upserting Variable %s@%s", variable.Path, variable.Namespace)
+	if _, _, err := client.Variables().Create(&variable, nil); err != nil {
+		return fmt.Errorf("error inserting variable %s@%s: %s", variable.Path, variable.Namespace, err.Error())
+	}
+	log.Printf("[DEBUG] Created variable %s@%s", variable.Path, variable.Namespace)
+
+	return resourceVariableRead(d, meta)
+}
+
+func resourceVariableDelete(d *schema.ResourceData, meta any) error {
+	client := meta.(ProviderConfig).client
+
+	path := d.Get("path").(string)
+	ns := d.Get("namespace").(string)
+	d.SetId(path + "@" + ns)
+
+	log.Printf("[DEBUG] Deleting variable %q", d.Id())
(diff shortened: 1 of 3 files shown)


nomad/resource_variable.go
```diff
@@ -0,0 +1,101 @@
...
+				Sensitive: true,
+			},
+		},
+	}
+}
+
+func resourceVariableWrite(d *schema.ResourceData, meta any) error {
+	client := meta.(ProviderConfig).client
+
+	variable := api.Variable{
```
lgfa29, line 49:
```suggestion
	variable := &api.Variable{
```

Then we don't need to derefence it below.

nomad/resource_variable.go
```diff
@@ -0,0 +1,101 @@
...
+		Namespace: d.Get("namespace").(string),
+		Path:      d.Get("path").(string),
+		Items:     make(map[string]string),
+	}
+	for name, value := range d.Get("items").(map[string]any) {
+		variable.Items[name] = value.(string)
+	}
+
+	log.Printf("[DEBUG] Upserting Variable %s@%s", variable.Path, variable.Namespace)
+	if _, _, err := client.Variables().Create(&variable, nil); err != nil {
```
lgfa29, line 59:
```suggestion
	if _, _, err := client.Variables().Create(variable, nil); err != nil {
```

nomad/resource_variable.go
```diff
@@ -0,0 +1,101 @@
...
+
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"path": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
```
lgfa29, line 30: I think we may be able to apply some of the validation rules here:
https://github.com/hashicorp/nomad/blob/c7c8bd4ce2fcd9f2cd13278043b7baf1f8dd4f47/nomad/structs/variables.go#L189-L213

More specifically:
* the path cannot be `nomad`.
* if the path starts with `nomad/` it can only be `nomad/jobs`, `nomad/jobs/*`, or `nomad/job-templates/*`.

A description would be good as well.

nomad/resource_variable.go
```diff
@@ -0,0 +1,101 @@
...
+	return nil
+}
+
+func resourceVariableRead(d *schema.ResourceData, meta any) error {
+	client := meta.(ProviderConfig).client
+
+	path := d.Get("path").(string)
+	ns := d.Get("namespace").(string)
+
+	log.Printf("[DEBUG] Reading variable %q", d.Id())
```
lgfa29, line 89: Is `d.Id()` set if the resource is new and is being fetched for the first time?

nomad/resource_variable.go
```diff
@@ -0,0 +1,101 @@
...
+		return fmt.Errorf("Failed to get information about %s@%s: %v", path, ns, err)
+	}
+	d.SetId(path + "@" + ns)
+	return d.Set("items", variable.Items)
+}
+
+func resourceVariableExists(d *schema.ResourceData, meta any) (bool, error) {
+	err := resourceVariableRead(d, meta)
+	return err == nil, nil
+}
```
lgfa29, line 101: I think we need if `err` is a 404? Kind of like here:
https://github.com/hashicorp/terraform-provider-nomad/blob/77c12b04db15efd9f7bc84d13f229e9bba861c1c/nomad/resource_acl_policy.go#L137-L142

lgfa29 (changes requested):
Thanks for the initial work @iluminae!

I think this is looking good, having tests and documentation are probably the main things missing.

Let us know if you need help with those 🙂

regner:
@iluminae will you be continuing this PR at all? This is a feature we are very much in need of, so would be willing to help if required.

regner:
I am going to see if I can take your work and get it over the line. A few things:
* What is the best practice for this sort of thing? I assume push to my own fork and open a new PR?
* I will probably respond to some comments since they wont be on my new PR
* I am relatively knew to Go, so this could be fun

nomad/resource_variable.go
```diff
@@ -0,0 +1,101 @@
...
+func resourceVariableWrite(d *schema.ResourceData, meta any) error {
+	client := meta.(ProviderConfig).client
+
+	variable := api.Variable{
+		Namespace: d.Get("namespace").(string),
+		Path:      d.Get("path").(string),
+		Items:     make(map[string]string),
+	}
+	for name, value := range d.Get("items").(map[string]any) {
+		variable.Items[name] = value.(string)
```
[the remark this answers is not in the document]
regner: Hey @lgfa29,

The namespace resource does the same by the looks of it:
```go
	for name, value := range d.Get("meta").(map[string]interface{}) {
		m[name] = value.(string)
	}
```

So it seems to me that if this was a concern it would be something that should be solved holistically for the provider.

nomad/resource_variable.go
regner: Forgive me if I misunderstand: As this is not currently a pointer, there is currently no "dereferencing" occurring below. Correct?

May I ask what the advantage of defining variable as a pointer here on line 49 is compared to what the original author did of passing it through as a pointer on line 59?

nomad/resource_variable.go
regner: Would it make sense to use `validation.StringMatch` with a regex of questionable complexity, or a custom validation function that does some simpler validation but is larger in code.

nomad/resource_variable.go
regner: Is the concern about it being fetched before it has been created?

regner: Right OK, just below that `d.SetId` sets the ID as the path + namespace.

iluminae:
ill just close this in favor of #325

PR #305, opened 2023-01-31
resource: new sso auth method binding rule resources

jrasell:
Raising as WIP for initial feedback. Deps will certainly need sorting before this is ready for merge.


nomad/resource_acl_auth_method.go
```diff
@@ -0,0 +1,407 @@
...
+			},
+			"token_locality": {
+				Description: `Defines whether the ACL Auth Method creates a local or global token when performing SSO login. This field must be set to either "local" or "global".`,
+				Required:    true,
+				Type:        schema.TypeString,
+				ValidateFunc: validation.StringInSlice([]string{
+					api.ACLAuthMethodTokenLocalityGlobal,
+					api.ACLAuthMethodTokenLocalityLocal,
+				}, false),
+			},
```
lgfa29, line 51: Is this equivalent to the ACL token [`global`](https://registry.terraform.io/providers/hashicorp/nomad/latest/docs/resources/acl_token#global) attribute? It may be nice to keep the name consistent if that's the case.

nomad/resource_acl_auth_method.go
```diff
@@ -0,0 +1,407 @@
...
+				Type:        schema.TypeString,
+				Required:    true,
+				Sensitive:   true,
+			},
+			"oidc_scopes": {
+				Description: "",
+				Type:        schema.TypeList,
+				Elem:        &schema.Schema{Type: schema.TypeString},
+				Optional:    true,
+			},
```
lgfa29, line 98: Would it make sense to have an `oidc` block instead of a flat config? So a config would look like this:

```hcl
resource "nomad_acl_auth_method" "okta" {
  name = "okta"
  
  oidc {
    discovery_url = "..."
    client_id     = "..."
  }
}
```

We may be able to even remove the `type` 🤔

nomad/resource_acl_auth_method.go
```diff
@@ -0,0 +1,407 @@
...
+	}
+
+	// Create our ACL auth method.
+	log.Print("[DEBUG] Creating ACL Auth Method")
+	aclBindingRuleCreateResp, _, err := client.ACLAuthMethods().Create(aclAuthMethod, nil)
+	if err != nil {
+		return fmt.Errorf("error creating ACL Auth Method: %s", err.Error())
+	}
+	log.Printf("[DEBUG] Created ACL Auth Method %q", aclBindingRuleCreateResp.Name)
+	d.SetId(aclBindingRuleCreateResp.Name)
```
lgfa29, line 155: Just double checking: is the name guaranteed to be unique?

nomad/resource_acl_auth_method.go
```diff
@@ -0,0 +1,407 @@
...
+func resourceACLAuthMethodUpdate(d *schema.ResourceData, meta interface{}) error {
+	providerConfig := meta.(ProviderConfig)
+	client := providerConfig.client
+
+	aclBindingRule, err := generateNomadACLAuthMethod(d)
+	if err != nil {
+		return err
+	}
+
+	log.Printf("[DEBUG] %#v", aclBindingRule.Config)
```
lgfa29, line 188: I think this may leak the sensitive values? I'm not sure if Terraform strips them from logs 🤔

nomad/resource_acl_auth_method.go
```diff
@@ -0,0 +1,407 @@
...
+		return fmt.Errorf("error reading ACL Auth Method %q: %s", authMethodName, err.Error())
+	}
+	log.Printf("[DEBUG] Read ACL Auth Method %q", authMethod.Name)
+
+	_ = d.Set("name", authMethod.Name)
+	_ = d.Set("type", authMethod.Type)
+	_ = d.Set("token_locality", authMethod.TokenLocality)
+	_ = d.Set("max_token_ttl", authMethod.MaxTokenTTL)
+	_ = d.Set("default", authMethod.Default)
+	_ = d.Set("config", flattenACLAuthMethodConfig(authMethod.Config))
```
lgfa29, line 225: There's a `StateWriter` helper to handle multiple `Set` like this, but we haven't use it much yet 😅 
https://github.com/hashicorp/terraform-provider-nomad/blob/main/nomad/data_source_scaling_policy.go#L74-L79

lgfa29 (commented):
Minor comments but looking good!

nomad/resource_acl_auth_method.go
jrasell: It is and this mirrors what we have in Nomad; seems wrong to change it here and rather keep the consistency with naming between the provider and Nomad API.

nomad/resource_acl_auth_method.go
jrasell: It's not a bad shout; I don't know how we feel about changing the layout compared to the API. I'll play around with changing it, when I have a moment.

nomad/resource_acl_auth_method.go
jrasell: Yes.

nomad/resource_acl_auth_method.go
jrasell: Oh yeah, good shout, we will.

nomad/resource_acl_auth_method.go
jrasell: I am inclined to keep this the same shape as the API and make things easier to translate between the two.

lgfa29 (approved):

jrasell:
@chrispruitt I hope to be working on this over the next days.

PR #314, opened 2023-03-22
Add workload Id and job ACLs

### LINKED ISSUE #306: Request: Support applying ACLs to workloads
Nomad has added a built in variables system. Along with that came the ability to manage access to variables with ACL policies. Nomad now has the ability to grant a specific job, group, or task an ACL.

This is a requirement for the tasks to be able to read variables if they are deployed to anything other than the default namespace.

https://developer.hashicorp.com/nomad/docs/concepts/workload-identity#workload-associated-acl-policies

This would require adding optional attributes for namepsace, job, group, and task to the nomad_acl_policy resource.

IamTheFij:
Adds support for passing ACLs for workloads when defining a policy.

I'm not sure how to write unit tests for this as the test section seemed like it was for acceptance testing. I wasn't able to get acceptance testing to run locally on my machine yet.

Fixes #306


nomad/resource_acl_policy.go
```diff
@@ -43,10 +44,78 @@ func resourceACLPolicy() *schema.Resource {
 				Required:    true,
 				Type:        schema.TypeString,
 			},
+
+			"workload_identity": {
```
lgfa29, line 48:
```suggestion
			"job_acl": {
```

It's better to keep the field name consistent with the API. The variables and function names below should be updated as well.

nomad/resource_acl_policy.go
```diff
@@ -43,10 +44,78 @@ func resourceACLPolicy() *schema.Resource {
...
+							Optional:    true,
+						},
+						"task": {
+							Description: "Task",
+							Type:        schema.TypeString,
+							Optional:    true,
+						},
+					},
+				},
+			},
```
lgfa29, line 78:
```suggestion
			},
			ValidateFunc: func(i interface{}, k string) ([]string, []error) {
				// ...
			},
```

There are some validations rules applied here that are static enough that I think would be worth checking at the provider so they are caught during plan:
https://github.com/hashicorp/nomad/blob/c7c8bd4ce2fcd9f2cd13278043b7baf1f8dd4f47/nomad/structs/structs.go#L12373-L12386

nomad/resource_acl_policy.go
```diff
@@ -43,10 +44,78 @@ func resourceACLPolicy() *schema.Resource {
...
 		},
 	}
 }
 
+func parseWorkloadIdentity(workloadIdentity interface{}) (*api.JobACL, error) {
+	jobPolicyList, ok := workloadIdentity.([]interface{})
+	if !ok || len(jobPolicyList) != 1 {
+		return nil, errors.New("failed to unpack workload_identity configuration block")
```
lgfa29, line 86:
```suggestion
		return nil, errors.New("only one workload_identity block allowed")
```

It's better to give specific guidance and context on what the error was and keep them different from the error message below so we know which code path was hit when debugging problems 😄

lgfa29 (changes requested):
Thanks for the PR @IamTheFij, and apologies for the delay in review it.

I left a few comments but it would also be nice to have some tests and documentation updates about this. What problems were you facing in running the tests locally?

I'm not sure if you've seen it, but we have some instructions in our README:
https://github.com/hashicorp/terraform-provider-nomad#developing-the-provider

For this work I think we can update `testResourceACLPolicy_initialConfig` to include some `JobACL` config:

https://github.com/hashicorp/terraform-provider-nomad/blob/77c12b04db15efd9f7bc84d13f229e9bba861c1c/nomad/resource_acl_policy_test.go#L114-L127

And then update `testResourceACLPolicy_initialCheck` to verify the job identity was set.

A more complete test would be to test the validation I mentioned. For this you can add a `ExpectError` to your [`TestStep`](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-sdk@v1.17.2/helper/resource#TestStep) to verify that the resource is invalid.

For the docs, we would need to update https://github.com/hashicorp/terraform-provider-nomad/blob/main/website/docs/r/acl_policy.html.markdown to include the new field. If you run `make website` you can preview your changes.

Let us know if you need help with any of this 🙂

nomad/resource_acl_policy.go
IamTheFij: I made this change but I get an error `ValidateFunc is not yet supported on lists or sets`. It's also unclear if there are plans or willingness to add support for this either: https://github.com/hashicorp/terraform-plugin-sdk/issues/156

IamTheFij: I'm instead moving all the validation to the parse function where I had some partial validation before.

IamTheFij:
Any advice on getting this test working so it can get across the finish line?

nomad/resource_acl_policy.go
```diff
@@ -43,10 +44,96 @@ func resourceACLPolicy() *schema.Resource {
 				Required:    true,
 				Type:        schema.TypeString,
 			},
+
+			"job_acl": {
+				Description: "Workload identity that this should be applied to.",
```
lgfa29, line 49:
```suggestion
				Description: "Workload identity association that should be applied to the policy.",
```

nomad/resource_acl_policy.go
```diff
@@ -124,6 +228,15 @@ func resourceACLPolicyRead(d *schema.ResourceData, meta interface{}) error {
 	d.Set("description", policy.Description)
 	d.Set("rules_hcl", policy.Rules)
 
+	if policy.JobACL != nil {
+		d.Set("job_acl", map[string]string{
+			"namespace": policy.JobACL.Namespace,
+			"job_id":    policy.JobACL.JobID,
+			"group":     policy.JobACL.Group,
+			"task":      policy.JobACL.Task,
+		})
```
lgfa29, line 237:
```suggestion
		d.Set("job_acl", []map[string]string{{
			"namespace": policy.JobACL.Namespace,
			"job_id":    policy.JobACL.JobID,
			"group":     policy.JobACL.Group,
			"task":      policy.JobACL.Task,
		}})
```

I think this is the problem. 

It's a little weird, but blocks in HCL are lists and, as the schema defines, we need a list of one element here. Calling `d.Set` with the wrong value type returns an error (which we inadvertently often ignore 😅) and so `job_acl` was never been written to state, causing the import operation to fail.

To log things while running test you can `log.Printf("[DEBUG] ...)` whatever you need and then set the env var `TF_LOG=DEBUG` while running the test.

lgfa29 (changes requested):
Looking good! Sorry for the delay in getting back to you here.

After the read method is fixed I think the only bits missing are updating the [documentation](https://github.com/hashicorp/terraform-provider-nomad/blob/main/website/docs/r/acl_policy.html.markdown) and adding a [CHANGELOG](https://github.com/hashicorp/terraform-provider-nomad/blob/main/CHANGELOG.md) entry. Something like:

```diff
## 1.4.21 (Unreleased)
+ IMPROVEMENTS:
+ * resource/nomad_acl_policy: add support for `job_acl` ([#314](https://github.com/hashicorp/terraform-provider-nomad/pull/314))

## 1.4.20 (April 20, 2023)
```

IamTheFij:
I haven't tested the website because the make target is broken. It executes `website-provider` on https://github.com/hashicorp/terraform-website, but that target doesn't exist there.

It looks like the target was added back in 2018 with [this PR](https://github.com/hashicorp/terraform-website/pull/233), but as of 2022 it was reportedly absent: https://github.com/hashicorp/terraform-provider-tfe/issues/503.

It looks like this is because the whole system was rewritten and the Makefile removed with this PR: https://github.com/hashicorp/terraform-website/pull/2010 and then added back without this ability in this PR: https://github.com/hashicorp/terraform-website/pull/2013

lgfa29 (approved):
Hum...yeah, that is weird. I'm not sure what happened with the website preview but I ran the changes through https://registry.terraform.io/tools/doc-preview and they look good. I pushed a commit with minor touch-ups.

I also realized that we may need to update the `nomad_acl_policy` data source, but I have bother you enough in this PR 😄 

PR #325, opened 2023-05-04
Adding support for nomad variables

regner:
This PR is based on #304. The original author, iluminae, of that PR is unable to continue their work.

Please note that I have not done Terraform provider work before, and very little Go work, so if some choices or questions don't make sense, that's probably why.

I have tried to implement the feedback that @lgfa29 had on the original PR. I did however only add a single test, I wanted to make sure I was on the write path before continuing.

### DIFF
--- a/nomad/resource_variable.go
+++ b/nomad/resource_variable.go
@@ -47,6 +47,7 @@ func resourceVariable() *schema.Resource {
 				Description: "Variable namespace",
 				Type:        schema.TypeString,
 				Optional:    true,
+				ForceNew:    true,
 				Default:     api.DefaultNamespace,
 			},
 			"items": {
@@ -153,12 +154,12 @@ func pathValidation() schema.SchemaValidateFunc {
 
 		// Limit length to 128 characters
 		if len(path) > maxPathLength {
-			errs = append(errs, fmt.Errorf("expecnted path legnth to be less than %v but got a path legnth of %v", maxPathLength, len(path)))
+			errs = append(errs, fmt.Errorf("expected path legnth to be less than %v but got a path legnth of %v", maxPathLength, len(path)))
 		}
 
 		// Limit to RFC3986 URL-safe characters, minus '@' and '.' as they conflict with template blocks
 		if !validVariablePath.MatchString(path) {
-			errs = append(errs, fmt.Errorf("the path %s contains invalid characters", path))
+			errs = append(errs, fmt.Errorf("path %s contains invalid characters", path))
 		}
 
 		// Validate paths that have 'nomad' as the first directory in the path
@@ -167,7 +168,7 @@ func pathValidation() schema.SchemaValidateFunc {
 		parts := strings.Split(path, "/")
 
 		if parts[0] != "nomad" {
-			return nil, nil
+			return nil, errs
 		}
 
 		if len(parts) == 1 {
@@ -176,9 +177,9 @@ func pathValidation() schema.SchemaValidateFunc {
 
 		switch {
 		case parts[1] == "jobs":
-			return nil, nil
+			return nil, errs
 		case parts[1] == "job-templates" && len(parts) == 3:
-			return nil, nil
+			return nil, errs
 		case parts[1] == "job-templates":
 			errs = append(errs, fmt.Errorf("the path 'nomad/job-templates' is reserved, you may write variables at the level below it, for example, 'nomad/job-templates/template-name'"))
 		default:
--- a/nomad/resource_variable.go
+++ b/nomad/resource_variable.go
@@ -91,7 +91,7 @@ func resourceVariableDelete(d *schema.ResourceData, meta any) error {
 
 	log.Printf("[DEBUG] Deleting variable %q", variableID)
 	if _, err := client.Variables().Delete(path, &api.WriteOptions{Namespace: ns}); err != nil {
-		return fmt.Errorf("Failed to delete Variable %s: %v", variableID, err)
+		return fmt.Errorf("error deleting variable %s: %v", variableID, err)
 	}
 
 	log.Printf("[DEBUG] Deleted Variable %q", d.Id())
@@ -103,22 +103,17 @@ func resourceVariableDelete(d *schema.ResourceData, meta any) error {
 func resourceVariableRead(d *schema.ResourceData, meta any) error {
 	client := meta.(ProviderConfig).client
 
-	variableID := d.Id()
-	// If the variable has not been created, the ID will be an empty string
-	// which means we can skip attempting to perform the lookup.
-	if variableID == "" {
-		return nil
-	}
-
 	path := d.Get("path").(string)
 	ns := d.Get("namespace").(string)
+	variableID := path + "@" + ns
 
 	log.Printf("[DEBUG] Reading variable %s", variableID)
 	variable, _, err := client.Variables().Read(path, &api.QueryOptions{Namespace: ns})
 	if err != nil {
-		return fmt.Errorf("Failed to get information about %s: %v", variableID, err)
+		return fmt.Errorf("error getting information about %s: %v", variableID, err)
 	}
 
+	d.SetId(variableID)
 	return d.Set("items", variable.Items)
 }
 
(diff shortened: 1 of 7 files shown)


regner:
@lgfa29 sorry for the ping, wasn't sure who else to poke, would you be able to help move this PR forward?

regner:
I am unsure how best to handle the path changing. That requires deleting the old resource and creating a new one.

nomad/resource_variable.go
```diff
@@ -0,0 +1,185 @@
...
+		Items:     make(map[string]string),
+	}
+
+	for name, value := range d.Get("items").(map[string]any) {
+		variable.Items[name] = value.(string)
+	}
+
+	log.Printf("[DEBUG] Upserting Variable %s@%s", variable.Path, variable.Namespace)
+	if _, _, err := client.Variables().Create(variable, nil); err != nil {
+		return fmt.Errorf("error creating Variable %s@%s: %s", variable.Path, variable.Namespace, err.Error())
```
lgfa29, line 76:
```suggestion
		return fmt.Errorf("error creating variable %s@%s: %s", variable.Path, variable.Namespace, err)
```

Just some minor nit-picking 😅

nomad/resource_variable.go
```diff
@@ -0,0 +1,185 @@
...
+
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"path": {
+				Description:  "The path at which the variable items will be stored, must be between 1 and 128 characters in length, be URL safe, and not include '@' or '.' characters",
+				Type:         schema.TypeString,
+				Required:     true,
```
lgfa29, line 42:
```suggestion
				Required:     true,
				ForceNew:     true,
```

I think updating the variable path requires a destructive update, otherwise we risk leaving stray variables when the path is changed.

nomad/resource_variable.go
```diff
@@ -0,0 +1,185 @@
...
+	variableID := d.Id()
+	path := d.Get("path").(string)
+	ns := d.Get("namespace").(string)
+
+	log.Printf("[DEBUG] Deleting variable %q", variableID)
+	if _, err := client.Variables().Delete(path, &api.WriteOptions{Namespace: ns}); err != nil {
+		return fmt.Errorf("error deleting variable %s: %v", variableID, err)
+	}
+
+	log.Printf("[DEBUG] Deleted Variable %q", d.Id())
```
lgfa29, line 97:
```suggestion
	log.Printf("[DEBUG] Deleted variable %q", d.Id())
```

nomad/resource_variable.go
```diff
@@ -0,0 +1,185 @@
...
+
+	_, _, err := client.Variables().Read(path, &api.QueryOptions{Namespace: ns})
+	if err != nil {
+		// As of Nomad 0.4.1, the API client returns an error for 404
+		// rather than a nil result, so we must check this way.
+		if strings.Contains(err.Error(), "404") {
+			return false, nil
+		}
+
+		return true, fmt.Errorf("error checking for Variable %s: %#v", variableID, err)
```
lgfa29, line 137:
```suggestion
		return true, fmt.Errorf("error checking for variable %s: %#v", variableID, err)
```

nomad/resource_variable.go
```diff
@@ -0,0 +1,185 @@
...
+			return nil, nil
+		case parts[1] == "job-templates" && len(parts) == 3:
+			return nil, nil
+		case parts[1] == "job-templates":
+			return nil, []error{fmt.Errorf("the path 'nomad/job-templates' is reserved, you may write variables at the level below it, for example, 'nomad/job-templates/template-name'")}
+		default:
+			return nil, []error{fmt.Errorf("only paths at 'nomad/jobs' or 'nomad/job-templates' and below are valid paths under the top-level 'nomad' directory")}
+		}
+	}
+}
```
lgfa29, line 185: Since Terrafor validation allows returning multiple errors, it's always best to check for and return as many errors as possible. This avoids the bad UX of users having to go through multiple cycles of `apply-error-fix-apply-error-fix-...`.

In this case I think only the first error is required to return early, so the function could look something like this:

```go
func pathValidation() schema.SchemaValidateFunc {
	return func(i interface{}, k string) ([]string, []error) {
		// Verify path actually is a string
		path, ok := i.(string)
		if !ok {
			return nil, []error{fmt.Errorf("expected type of %s to be string", k)}
		}

		errs := []error{}

		// Limit length to 128 characters
		if len(path) > maxPathLength {
			errs = append(errs, fmt.Errorf("expecnted path legnth to be less than %v but got a path legnth of %v", maxPathLength, len(path)))
		}

		// Limit to RFC3986 URL-safe characters, minus '@' and '.' as they conflict with template blocks
		if !validVariablePath.MatchString(path) {
			errs = append(errs, fmt.Errorf("the path %s contains invalid characters", path))
		}

		// Validate paths that have 'nomad' as the first directory in the path
		// The following validation is taken from:
		// https://github.com/hashicorp/nomad/blob/c7c8bd4ce2fcd9f2cd13278043b7baf1f8dd4f47/nomad/structs/variables.go#L191
		parts := strings.Split(path, "/")

		if parts[0] != "nomad" {
			return nil, errs
		}

		if len(parts) == 1 {
			errs = append(errs, fmt.Errorf("path of 'nomad' is a reserved top-level directory path"))
		}

		switch {
		case parts[1] == "jobs":
			return nil, errs
		case parts[1] == "job-templates" && len(parts) == 3:
			return nil, errs
		case parts[1] == "job-templates":
			errs = append(errs, fmt.Errorf("the path 'nomad/job-templates' is reserved, you may write variables at the level below it, for example, 'nomad/job-templates/template-name'"))
		default:
			errs = append(errs, fmt.Errorf("only paths at 'nomad/jobs' or 'nomad/job-templates' and below are valid paths under the top-level 'nomad' directory"))
		}
		return nil, errs
	}
}
```

nomad/resource_variable_test.go
lgfa29, line 1: It would nice to have an update test as well, I think it may have caught the missing `ForceNew`. Here's an example for namespace:
https://github.com/hashicorp/terraform-provider-nomad/blob/f30b1f6d855fab94bc1278ab8825b7ec5122ee73/nomad/resource_namespace_test.go#L77-L114

Let me know if you need any extra help on this 🙂

lgfa29 (changes requested):
Looking good @regner, thanks for picking-up the work from @iluminae.

I left a few more notes here, and would you be able to write some documentation as well? They live inside the `website` forlder.  Unfortunately the local doc preview flow is not working anymore, but you can use this website for now: https://registry.terraform.io/tools/doc-preview

Let me know if you have any questions 🙂

nomad/resource_variable.go
regner: I did not realize I could return multiple errors like that. Thank you very much for pointing that out. It would indeed make it might nicer to get a list of errors at once.

nomad/resource_variable_test.go
regner: I added another test to validate the variable path changing. Having looked at that namespace test, from what I can tell it doesn't validate that the old namespace is deleted. I added a `CheckDestroy` function on the path change test, and it seems like the namespace one should have that as well.

regner:
Thank you very much for the comments. I have updated everything but the docs, which I will work on now.

regner:
@lgfa29 I have now added the docs. I think it's ready for another review when you have time. Thank you very much for your time and help with this.

regner:
Is there a way to run a specific test or do I need to always run all the tests?

lgfa29 (approved):
Thank you for the fixes @regner!

I noticed some other things I didn't notice in the first review (like namespace needing to be `ForceNew` as well) and others mistake I made with the error handling where some errors ended up being ignore. I pushed https://github.com/hashicorp/terraform-provider-nomad/pull/325/commits/2a9a0c08439d783c1f281d68b47871d94ad85969 to fix those and some other minor things. Hopefully I didn't break anything else 😅

lgfa29:
> Is there a way to run a specific test or do I need to always run all the tests?

For this I usually run the `go test` command by hand, like 

```
TF_ACC=1 go test -count 1 -v -run TestResourceVariable ./nomad
```

nomad/resource_variable_test.go
lgfa29: Yeah, that's a good catch!

PR #326, opened 2023-05-11
Fix make localtestacc

IamTheFij:
There are a few layered bugs that this fixes.

1.  When trying to capture the output of `scripts/start-nomad.sh`, the `shell` function never exits. I believe it's waiting on all processes spawned to exist, but we don't kill them until after the test. So the test execution was hanging. This was resolved by avoiding capture and instead reading the token from a file.
2. Apparently the shell function substitution doesn't happen sequentially within the target. When running the script and trying reading the variable using `$(shell cat /tmp/nomad-test.token)` right after, an error that `/tmp/nomad-test.token` doesn't exist is echoed before the script is even run. In order to ensure the script is run before trying to read the token, this is moved to a different target.
3. Output from the running nomad was being echoed to the screen despite the process being in the background. This is resolved by redirecting the output.
4. Running `stop-nomad.sh` stopped consul and vault, but not nomad. This is because `$! was actually capturing the pid of consul a second time because nomad was backgrounded by `sudo` and not the current shell. This was resolved by having sudo execute a bash shell and have that spawned shell capture the pid instead.
5. Capturing output as `env NOMAD_TOKEN=$(./scripts/star-nomad.sh)` wouldn't have worked anyway because it was actually echoing `NOMAD_TOKEN=<THETOKEN>` as well as a bunch of stdout from nomad commands. This was resolved by conditionally echoing the token alone and extraneous stdout to stderr.
6. Finally, the pid files were being deleted even if the kill commands were not successful. This could have been resolved by adding `set -e`, but I refactored it a bit more to kill whatever it could but allow retrying or further investigation for processes it couldn't kill.

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -16,10 +16,13 @@ test: fmtcheck
 testacc: fmtcheck
 	TF_ACC=1 go test $(TEST) -v $(TESTARGS) -timeout 120m -count=1
 
-localtestacc: fmtcheck
-	-export NOMAD_TOKEN=$(shell scripts/start-nomad.sh); \
-	export TF_ACC=1; \
-	go test $(TEST) -v $(TESTARGS) -timeout 120m -count=1
+localtestacc-start-nomad:
+	scripts/start-nomad.sh
+
+localtestacc: fmtcheck localtestacc-start-nomad
+	-env NOMAD_TOKEN=$(shell cat /tmp/nomad-test.token) \
+		TF_ACC=1 \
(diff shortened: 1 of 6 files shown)


lgfa29 (approved):
Thanks for the fixed @IamTheFij! 

I took the opportunity to also update the script to use the new bootstrap with fixed token functionality (https://developer.hashicorp.com/nomad/docs/commands/acl/bootstrap#examples) which greatly simplifies the Nomad token management.

I also updated the version of Consul, Nomad, and Vault being installed because they were pretty outdated 😅

PR #330, opened 2023-06-05
making bind_name optional when bind_type is management

### LINKED ISSUE #324: nomad_acl_binding_rule requires bind_name when it should not
### Nomad Version
1.5.3

### Affected Resource(s)
- nomad_acl_binding_rule

```hcl
resource "nomad_acl_binding_rule" "main" {
  description = "example management rule"
  auth_method = nomad_acl_auth_method.main.name
  selector    = "admin in list.roles"
  bind_type   = "management"
}
```

### Expected Behavior
This should create an ACL binding rule that binds to a management token.

### Actual Behavior
It fails due to requiring the bind_name argument.

### Important Factoids
The Nomad documentation specifies the bind-name argument as optional, and specifically says it should not be provided when the bind type is management.

> [-bind-name](https://developer.hashicorp.com/nomad/docs/commands/acl/binding-rule/create#bind-name): Specifies is the target of the binding used on selector match. This can be lightly templated using HIL ${foo} syntax. If the bind type is set to management, this should not be set.

https://developer.hashicorp.com/nomad/docs/commands/acl/binding-rule/create

Of note, providing an empty string (`""`) does work to get around this.

regner:
When creating a ACL binding rule, the bind name must not be defined but only when the bind_type is set to management.

https://developer.hashicorp.com/nomad/docs/commands/acl/binding-rule/create

Fixes #324

### DIFF
--- a/nomad/resource_acl_binding_rule.go
+++ b/nomad/resource_acl_binding_rule.go
@@ -46,14 +46,14 @@ func resourceACLBindingRule() *schema.Resource {
 				Required:    true,
 				Type:        schema.TypeString,
 				ValidateFunc: validation.StringInSlice([]string{
+					api.ACLBindingRuleBindTypeManagement,
 					api.ACLBindingRuleBindTypePolicy,
 					api.ACLBindingRuleBindTypeRole,
-					"management",
 				}, false),
 			},
 			"bind_name": {
 				Description: "Target of the binding.",
-				Required:    true,
+				Optional:    true,
 				Type:        schema.TypeString,
 			},
 		},
@@ -61,6 +61,11 @@ func resourceACLBindingRule() *schema.Resource {
 }
 
 func resourceACLBindingRuleCreate(d *schema.ResourceData, meta interface{}) error {
+	err := validateNomadACLBindingRule(d)
+	if err != nil {
+		return err
+	}
+
 	providerConfig := meta.(ProviderConfig)
 	client := providerConfig.client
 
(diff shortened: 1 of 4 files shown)


pkazmierczak (approved):
Hey @regner, thanks for the PR!
LGTM!

nomad/resource_acl_binding_rule.go
```diff
@@ -172,3 +182,18 @@ func generateNomadACLBindingRule(d *schema.ResourceData) *api.ACLBindingRule {
...
+	bindName := d.Get("bind_name").(string)
+	bindType := d.Get("bind_type").(string)
+
+	if bindType != api.ACLBindingRuleBindTypeManagement && bindName == "" {
+		return fmt.Errorf("error bind_name must be defined if bind_type is not '%q'", api.ACLBindingRuleBindTypeManagement)
+	}
+
+	if bindType == api.ACLBindingRuleBindTypeManagement && bindName != "" {
+		return fmt.Errorf("error bind_name must not be defined if bind_type is '%q'", api.ACLBindingRuleBindTypeManagement)
+	}
```
lgfa29, line 196: I think this is a bit easier to read?

```go
switch bindType {
case api.ACLBindingRuleBindTypeManagement:
	if bindName != "" {
		return fmt.Errorf("error bind_name must not be defined if bind_type is '%q'", api.ACLBindingRuleBindTypeManagement)
	}
default:
	if bindName == "" {
		return fmt.Errorf("error bind_name must be defined if bind_type is '%q'", bindType)
	}
}
```

lgfa29 (approved):
Minor nit-picking in the validation check, but not a blocker. Thanks for fixing this! 🙂

regner:
Adjusted the validateNomadACLBindingRule formatting based on @lgfa29 feedback. I am definitely not used to reaching for switch statements so they don't come to my mind first, and I personally find it a bit harder to read, but I think that's just because I so rarely use them.

Thank you for the suggestion. I can definitely see how it would be easier to read once I train my brain a bit more.

Issue #321, opened 2023-04-26
nomad_quota_specification is missing support for certain parameters


henrikjohansen:
`nomad quota init` yields the following content for _Nomad v1.5.3_ : 

```
name        = "default-quota"
description = "Limit the shared default namespace"

# Create a limit for the global region. Additional limits may
# be specified in-order to limit other regions.
limit {
  region = "global"
  region_limit {
    cpu        = 2500
    memory     = 1000
    memory_max = 1000
  }
  variables_limit = 1000
}
```

Currently, `memory_max` and `variables_limit` are not implemented by the nomad_quota_specification resource.
Support for the aforementioned would be highly appreciated :smile:

lgfa29:
Thanks for the report @henrikjohansen. Our quote resource does seem quite a bit out of date 😅

PR #332, opened 2023-06-06
Adding new attributes to the quota spec resource

regner:
Fix for #321. Just a first pass adding the attributes.

Still need to add tests and docs.

### DIFF
--- a/nomad/resource_quota_specification.go
+++ b/nomad/resource_quota_specification.go
@@ -64,8 +64,9 @@ func resourceQuotaSpecificationLimits() *schema.Resource {
 				Elem:        resourceQuotaSpecificationRegionLimits(),
 			},
 			"variables_limit": {
-				Type:     schema.TypeInt,
-				Optional: true,
+				Description: "The maximum total size of all variables.",
+				Type:        schema.TypeInt,
+				Optional:    true,
 			},
 		},
 	}
--- a/nomad/resource_quota_specification.go
+++ b/nomad/resource_quota_specification.go
@@ -63,6 +63,10 @@ func resourceQuotaSpecificationLimits() *schema.Resource {
 				MaxItems:    1,
 				Elem:        resourceQuotaSpecificationRegionLimits(),
 			},
+			"variables_limit": {
+				Type:     schema.TypeInt,
+				Optional: true,
+			},
 		},
 	}
 }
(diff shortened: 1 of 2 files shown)


regner:
@lgfa29 sorry to poke you directly, but question: with changes like this where an API is changing a lot are there concerns about backwards compatibility? It is all new attributes, nothing renamed or removed. So I assume the actual API calls should work.

regner:
Also for reference, here is the go struct for QuotaLimit that the API expects: https://pkg.go.dev/github.com/hashicorp/nomad/api#QuotaLimit

lgfa29:
> It is all new attributes, nothing renamed or removed. So I assume the actual API calls should work.

Yup, adding new attributes should be fine 👍

regner:
So problem: I am unable to run the tests for the quota specification as that is an enterprise feature, I don't have an enterprise license, and the install script (https://github.com/hashicorp/terraform-provider-nomad/blob/main/scripts/getnomad.sh#L8) does not install an enterprise version unless you have a license.

Should I just write the tests and hope they pass? Or is there a work around for this?

regner:
Question: What is the default value of optional integers?

Example existing code:
```
"cpu": {
  Type:     schema.TypeInt,
  Optional: true,
},
```

And the docs for that:
```
- `cpu` `(int: 0)` - The amount of CPU to limit allocations to. A value of zero
  is treated as unlimited, and a negative value is treated as fully disallowed.
```

What makes the default 0? Is the provider setting it to 0 somewhere, is that because the Go language treats it as 0, or is that because we know that the Nomad API will default it to 0 if the attribute is not set at all?

lgfa29:
Hi @regner, sorry for the delay getting back to you here. I had to focus on Nomad 1.6 so I ended up missing some of the other work streams.

> Should I just write the tests and hope they pass? Or is there a work around for this?

Yeah, that's not great, but I'm not sure if there's another solution available. For some reason CI is also not running tests on this PR (I will also look into this), so for now feel free to write tests and hope they work. I can run them manually before merge.

> What is the default value of optional integers?

You got it right here: "because the Go language treats it as 0".

> Make this slightly more confusing, some of the documentation will list optional strings differently:

Normally we mark it fields as `<optional>` when the zero-value does not impact the resource. In your example, `capacity_min` zero-value is an empty string `""` (again, because of Go) and this value has no special meaning for the resource. For ACL token `name`, the zero-value may be significant because you will usually see the token name in outputs etc.

Now, that's not an exact science and I'm sure there are plenty of inconsistencies 😅

regner:
Ah! Unsure how I missed the PR comments back then, but I do appreciate the new PR. Thanks for responding here to let me know.

PR #333, opened 2023-06-06
ci: fix tests

lgfa29:
I accidentally removed this line to set `NOMAD_TOKEN` to the runner environment in #326
https://github.com/hashicorp/terraform-provider-nomad/pull/326/commits/98e16b2ec4fca80a8f233e5f916dab980d725f8d#diff-a2adceb1476caa3f0076774cc3848b480976990a98d90b9fec9254a8a61a8bceL63

I also bumped Consul and Vault in that PR, but these newer enterprise versions require a license, but it turns out we don't actually need the ENT version of those tools in our tests, just Nomad.

### DIFF
--- a/scripts/getnomad.sh
+++ b/scripts/getnomad.sh
@@ -5,7 +5,7 @@
 
 set -e
 
-NOMAD_VERSION=1.5.6
+NOMAD_VERSION=1.5.0
 if [[ -n "$NOMAD_LICENSE" || -n "$NOMAD_LICENSE_PATH" ]]; then
     NOMAD_VERSION=${NOMAD_VERSION}+ent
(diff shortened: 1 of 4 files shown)


scripts/getnomad.sh
```diff
@@ -5,7 +5,7 @@
 
 set -e
 
-NOMAD_VERSION=1.5.6
+NOMAD_VERSION=1.5.0
```
lgfa29, line 8: Bumping past `1.5.1` causes `TestResourceJob_serviceDeploymentInfo` to fail. I'm still investigating why.

pkazmierczak (approved):

PR #339, opened 2023-07-07
deps: update Terraform SDK to v2

lgfa29:
The Terraform SDK v1 has long been unsupported. Upgrading to v2 means dropping support for Terraform < 0.12, which has also been unsupported for a while. 

This is a breaking change, so the next provider release will be v2.0.0.


jrasell (approved):
LGTM!

PR #340, opened 2023-07-07
Add resource and data sources for node pools

lgfa29:
Nomad dependency will need to be updated after GA, but implementation is ready for review. 

Targets `upgrade-sdkv2` branch to solve a [bug](https://github.com/hashicorp/terraform/issues/20487) I hit in tests using SDK v1.


nomad/resource_node_pool.go
```diff
@@ -0,0 +1,254 @@
...
+func resourceNodePool() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceNodePoolWrite,
+		Update: resourceNodePoolWrite,
+		Delete: resourceNodePoolDelete,
+		Read:   resourceNodePoolRead,
+		Exists: resourceNodePoolExists,
+
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
```
jrasell, line 33: This is deprecated in favour of `ImportStatePassthroughContext`, do we want to use the non-deprecated function for this initial commit?

nomad/resource_node_pool.go
```diff
@@ -0,0 +1,254 @@
...
+		},
+
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Description: "Unique name for this node pool.",
+				Type:        schema.TypeString,
+				Required:    true,
+				ForceNew:    true,
+			},
+			"description": {
```
jrasell, line 43: Do we want to add validation via `validation.StringLenBetween` here to avoid API calls that will just fail?

jrasell (approved):
LGTM with a couple of minor comments.

nomad/resource_node_pool.go
lgfa29: Oh nice, I didn't notice the deprecation note, thanks!

nomad/resource_node_pool.go
lgfa29: Added in https://github.com/hashicorp/terraform-provider-nomad/pull/340/commits/a5fcae001420ded22f00fc1431c77903a116775e

Issue #338, opened 2023-07-07
resource_volume doesn't detect topology_request drift


the-nando:
### Terraform Version
```
Terraform v1.4.6
on darwin_arm64
```

### Nomad Version
```
Nomad v1.5.6
```

### Provider Configuration
Which values are you setting in the provider configuration? 

```hcl
provider "nomad" {
  address = "${local.nomad_http_address}"
  region = "${local.nomad_region}"
  ignore_env_vars = {
    NOMAD_NAMESPACE = true
    NOMAD_REGION = true
  }
}
```

### Environment Variables
No

### Affected Resource(s)
Please list the resources as a list, for example:
- resource_volume

### Terraform Configuration Files
```hcl
resource "nomad_volume" "default_topology" {
  name        = "foo[0]"
  type        = "csi"
  plugin_id   = "aws-ebs"
  volume_id   = "foo[0]"
  external_id = <aws_ebs_id>
  namespace   = "test"

  capability {
    access_mode     = "single-node-writer"
    attachment_mode = "file-system"
  }

  mount_options {
    fs_type = "ext4"
  }

  topology_request {
    required {
      topology {
        segments = {
          "topology.ebs.csi.aws.com/zone" = each.value["availability_zone"]
        }
      }
    }
  }
}
```

If I create a Nomad volume with the code above, deregister the volume and recreate it without `topology_request`, Terraform won't notice the drift.

### Expected Behavior
Terraform configures the `topology_request` on the volume.

### Actual Behavior
```
No changes. Your infrastructure matches the configuration.
```

### Steps to Reproduce
1. `terraform apply`
2. `nomad volume deregister -namespace='test' 'foo[0]' `
3. Reregister the volume without the `topology_request` block:
```
~ cat > volume.hcl <<EOF                                                                 
id           = "foo[0]"
name         = "foo[0]"
namespace    = "monitoring"

type         = "csi"
plugin_id    = "aws-ebs"
external_id  = "vol-<xxx>"

capability {
  access_mode     = "single-node-writer"
  attachment_mode = "file-system"
}

mount_options {
  fs_type     = "ext4"
}
EOF

~ cat volume.hcl | nomad volume register -namespace "test" -
```
4. `terraform apply`

[...]

PR #342, opened 2023-07-08
resource_volume : set topology_request on read

the-nando:
Fixes: https://github.com/hashicorp/terraform-provider-nomad/issues/338

The MR covers only `required` topology_request as that's what's implemented in the schema for this resource. Should it be extended to account also for `preferred`? According the the [Volume Specification](https://developer.hashicorp.com/nomad/docs/other-specifications/volume) they are both supported.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ BUG FIXES:
 * data source/namespace: use type list to represent capabilities so its values can be indexed with Terraform SDKv2 ([#339](https://github.com/hashicorp/terraform-provider-nomad/issues/339))
 * data source/nomad_volume: fix panic when reading volume ([#323](https://github.com/hashicorp/terraform-provider-nomad/pull/323))
 * resources/nomad_acl_binding_rule: fix a bug where `bind_name` was required even when `bind_type` was `management`. ([#330](https://github.com/hashicorp/terraform-provider-nomad/pull/330))
+* resource/nomad_volume and resource/nomad_external_volume: fix a bug where `topology_request` was not persisted to state. ([#342](https://github.com/hashicorp/terraform-provider-nomad/pull/342)
 
 ## 1.4.20 (April 20, 2023)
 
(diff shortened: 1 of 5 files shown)


lgfa29:
Thanks for the PR @the-nando!

`nomad_volume` and `nomad_external_volume` are, due to historical reasons, two weird resources 😅 

As noted in the [docs](https://developer.hashicorp.com/nomad/docs/other-specifications/volume/topology_request#preferred) and in the [comment in the code](https://github.com/hashicorp/terraform-provider-nomad/blob/main/nomad/resource_volume.go#L213-L214) (though the URL is broken, so we need to fix that), `preferred` is only supported on volume **creation**, which `nomad_volume` does not do, only `nomad_external_volume` creates  volumes.

But the `Read` method is shared between them, so I think we do need to handle `preferred` so `nomad_external_volume` is fully fixed (and maybe add `ForceNew` there as well). Hopefully Terraform won't complain about the missing field in the schema, though I do think it might 😬 

If it does complain, then I think we may need to add it to the schema but as computed so users don't set it.

the-nando:
Hey @lgfa29, thanks for the clarification regarding the setting of `preferred` in `nomad_volumes` vs `nomad_external_volume` :)

Feel free to bundle up the changes in this PR with the ones in #344 and close this if that's easier for you. I'll wait then for the next major.

Thanks!

lgfa29:
Looking at it again, #344 was already big and convoluted enough that adding more changes to it would make it even more confusing 😅 

So I think I will rebase your branch and keep this PR up for this specific bug fix.

lgfa29 (approved):
I've updated the topology request flattening function to also read `preferred` and Terraform didn't seem to complain about it, so I didn't have to update the schema. I also noticed that `ForceNew` didn't work unless it was set in all levels of the schema, so I addedthat as well.

Lastly I rebased your branch to apply the fix to the new `nomad_csi_volume` and `nomad_csi_volume_request` resources, so i think we're all good here.

Free to reach out if you notice a problem with these changes.

Thanks again for the PR!

PR #343, opened 2023-07-10
resource/nomad_job: switch to HCL2 by default

lgfa29:
Parsing with HCL2 is the default behaviour of the Nomad CLI and so this change aims to make their behaviour the same.

The new `hcl1` attribute is provided for backwards compatibility to also match the Nomad CLI, which has the `-hcl1` flag.

### DIFF
--- a/nomad/resource_job.go
+++ b/nomad/resource_job.go
@@ -91,7 +91,7 @@ func resourceJob() *schema.Resource {
 					Schema: map[string]*schema.Schema{
 						"enabled": {
 							Description: "If true, the `jobspec` will be parsed as HCL2 instead of HCL.",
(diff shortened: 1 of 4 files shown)


nomad/resource_job.go
```diff
@@ -91,6 +91,7 @@ func resourceJob() *schema.Resource {
 					Schema: map[string]*schema.Schema{
 						"enabled": {
 							Description: "If true, the `jobspec` will be parsed as HCL2 instead of HCL.",
+							Deprecated:  "Starting with version 2.0.0 of the Nomad provider, jobs are parsed using HCL2 by default so this field is no longer used and may be safely removed from your configuration files. Set 'hcl1 = true' if you must use HCL1 job parsing.",
```
jrasell, line 94:
```suggestion
							Deprecated:  "Starting with version 2.0.0 of the Nomad provider, jobs are parsed using HCL2 by default, so this field is no longer used and may be safely removed from your configuration files. Set 'hcl1 = true' if you must use HCL1 job parsing.",
```

jrasell (approved):
LGTM!

PR #344, opened 2023-07-11
volumes: rename existing volume resources

lgfa29:
Rename `nomad_volume` to `nomad_csi_volume_registration` and `nomad_external_volume` to `nomad_csi_volume` to better reflect their intended use.

The deprecated top-level fields `access_mode` and `attachment_mode` were not moved to the new resource.

The previous resources are now marked as deprecated.

It's kind of hard to spot the differences between the original and new files, so here's a diff of them:
https://gist.github.com/lgfa29/c2a0c588da8724c3fd916b3e97f98b4b


jrasell (approved):
LGTM!

PR #345, opened 2023-07-12
deps: update Terraform SDK and Nomad

lgfa29:
I'm not sure how but I ended up picking an old version of the SDK in https://github.com/hashicorp/terraform-provider-nomad/pull/339 😅 

Also update the Nomad dependency to RC to pin the HCLv2 dependency to the same as Nomad, otherwise there's an import conflict with the SDK.

```
go install
# github.com/hashicorp/nomad/jobspec2
../../../../pkg/mod/github.com/hashicorp/nomad@v1.5.2/jobspec2/hcl_conversions.go:18:23: undefined: gohcl.Decoder
../../../../pkg/mod/github.com/hashicorp/nomad@v1.5.2/jobspec2/hcl_conversions.go:26:29: undefined: gohcl.Decoder
../../../../pkg/mod/github.com/hashicorp/nomad@v1.5.2/jobspec2/hcl_conversions.go:27:20: undefined: gohcl.Decoder
../../../../pkg/mod/github.com/hashicorp/nomad@v1.5.2/jobspec2/types.config.go:311:3: unknown field UndefinedVariable in struct literal of type hcl.EvalContext
```

The new SDK also picked some existing errors that were previously ignored in v1, so I'm fixing them here as well.


shoenig (approved):
LGTM!

Issue #300, opened 2022-12-16
Missing "namespace" in HTTP request in nomad_job with "detach=false" for ACL enabled Nomad


dezpriest:
### Description
We have an ACL enabled Nomad cluster. And I'm facing issues with using `nomad_job` resource with `detach=false` option.
I checked the Nomad server logs during the `terraform apply`. And found out that tf is querying endpoint `/v1/job/bonus-core/allocations?all=false`

This fits the [Nomad Job API documentation](https://developer.hashicorp.com/nomad/api-docs/jobs#list-job-allocations) ,but unfortunately returns `Permission denied` in our case. 

But if we add `?namespace=bonus-core` (`/v1/job/bonus-core/allocations?namespace=bonus-core`) into resulting query - we will get successfully all allocations for related job.

Seems like this point is missing in documentation. But this parameter mentioned for another [endpoint](https://developer.hashicorp.com/nomad/api-docs/jobs#namespace)
Query with `?namespace=*` sadly returns empty array.

### Terraform Version
`1.3.2`

### Nomad Version
`1.4.2 ACL enabled`

### Provider Configuration
```
provider "nomad" {
  address = "https://nomad.${terraform.workspace}.company.com:4646/"
  secret_id = data.vault_generic_secret.bonus-core-deploy.data.secret_id
}

data "vault_generic_secret" "bonus-core-deploy" {
  path = "nomad/creds/bonus-core-deploy"
}
```

### Affected Resource(s)
Please list the resources as a list, for example:
- nomad_job

### Terraform Configuration Files
```
resource "nomad_job" "core" {
    jobspec = file("specs/core.hcl")
    detach  = false
    hcl2 { 
        enabled = true
        vars = {
            "tf_vars" = local.tf_vars
            "image_id" = var.build_version
        }
    }
  }
```

ACL Policy (permissions less strict than usual for debug reasons)
```
namespace "bonus-core" {
  policy = "write"
}

node {
  policy = "write"
}

agent {
  policy = "write"
}

operator {
  policy = "write"
}

plugin {
  policy = "list"
}
```

### Debug Output
Output from Nomad Server logs
```
2022-12-16T15:10:18.319Z [DEBUG] http: request failed: method=GET path=/v1/job/bonus-core/allocations?all=false error="Permission denied" code=403
2022-12-16T15:10:18.319Z [DEBUG] http: request complete: method=GET path=/v1/job/bonus-core/allocations?all=false duration="341.456µs"
```

[...]

PR #347, opened 2023-07-12
resource/job: set namespace on monitor

lgfa29:
When monitoring the job deployment and allocations the provider needs to set the job namespace in the request instead of relying on the provider configuration as it may be different from the job.

Closes https://github.com/hashicorp/terraform-provider-nomad/issues/300


jrasell (approved):
LGTM!

I wonder if we want to add the namespace to any of the debug logs? Not required in this PR, but could be a followup if we think it's worth it.

lgfa29:
> I wonder if we want to add the namespace to any of the debug logs?

Added 👍

PR #348, opened 2023-07-13
ci: add copywrite workflow action to check headers

jrasell:


.github/workflows/copywrite.yml
```diff
@@ -0,0 +1,19 @@
...
+
+on:
+  push: {}
+
+jobs:
+  copywrite:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@8e5e7e5ab8b370d6c329ec480221332ada57f0ab # v3.5.2
+      - uses: hashicorp/setup-copywrite@867a1a2a064a0626db322392806428f7dc59cb3e # v1.1.2
```
lgfa29, line 11:
```suggestion
      - uses: hashicorp/setup-copywrite@v1
```

IIRC we're allowed to use versions for our actions.

lgfa29 (approved):

PR #357, opened 2023-07-26
resource/job: remove attribute allocation_ids

lgfa29:
Storing the list of allocation IDs in a `nomad_job` resource can lead to ever-changing plans since the list can change outsite of Terraform.

Retrieving this type of information is better suited for a data source.

The Terraform provider deprecation guide recommends attributes to be first marked as deprecated before full removal in the next major release.

Given that the next release will be a major release, this commit introduces a breaking change in behaviour (allocation IDs are not fetched by default anymore) but provides a fallback with the `read_allocation_ids` attribute.

Both attributes are marked as deprecated and will be removed in the next major release.

`nomad_allocations` data source added in https://github.com/hashicorp/terraform-provider-nomad/pull/358


gulducat (approved):
I like the availability of the fallback for now.  I assume you've tested it out of band, since the only test code changes are removals?

lgfa29:
Yup. This option causes several tests to be flaky and, since it will be removed in the future, I didn't think it was worth to keep a flaky test around.

PR #358, opened 2023-07-26
data source/allocations: add new data source

lgfa29:
Add new data source to retrieve a list of allocations. Support prefix and filter to restrict results.


pkazmierczak (approved):
LGTM!

nomad/data_source_allocations.go
```diff
@@ -0,0 +1,160 @@
+package nomad
```
pkazmierczak, line 1: I think you'll need copyright headers for each of the new files.

Issue #137, opened 2020-08-06
Add importer to resources


lgfa29:
Terraform allows importing existing resources into a state using the [`terraform import`](https://www.terraform.io/docs/import/index.html) command. In order to support this functionality resources must implement the functionalities described in this [guide](https://www.terraform.io/docs/extend/resources/import.html).

It may not be possible to add `Importer`s to all resources, but we should try adding them whenever possible. 

Here's a list of the current resources, we can check off when an importer is implemented or deemed not possible.

* [x] nomad_acl_policy
* [x] nomad_acl_token
* [ ] nomad_job
* [x] nomad_namespace
* [x] nomad_quota_specification
* [x] nomad_sentinel_policy
* [ ] nomad_volume

evandam:
Hey @lgfa29, sorry to bump an ancient issue, but do you think this would be possible at some point? I was considering switching some custom scripts and configuration management to use the Nomad provider but being able to import resources would be pretty important in getting started.

ACL policies, tokens, and namespaces would probably be at the top of our list for priority. I wish I could attempt this but I don't write Go 😅

lgfa29:
It's been a while huh 😞 

We currently don't have it a date to get this work in, but I also just realized that the list is wrong 😅 

I probably just listed all resources we had at the time without checking, but most of them are actually already importable, including the ones you listed. 

```hcl
provider "nomad" {
  address = "http://localhost:4646"
}

resource "nomad_acl_token" "bootstrap" {
  name   = "Bootstrap Token"
  type   = "management"
  global = true
}

resource "nomad_acl_policy" "test" {
  name      = "test"
  rules_hcl = file("${path.module}/policy.hcl")
}

resource "nomad_namespace" "dev" {
  name = "dev"
}
```

```console
$ terraform import nomad_namespace.dev dev
nomad_namespace.dev: Importing from ID "dev"...
nomad_namespace.dev: Import prepared!
  Prepared nomad_namespace for import
nomad_namespace.dev: Refreshing state... [id=dev]

Import successful!

The resources that were imported are shown above. These resources are now in
your Terraform state and will henceforth be managed by Terraform.

$ terraform import nomad_acl_policy.test test
nomad_acl_policy.test: Importing from ID "test"...
nomad_acl_policy.test: Import prepared!
  Prepared nomad_acl_policy for import
nomad_acl_policy.test: Refreshing state... [id=test]

Import successful!

The resources that were imported are shown above. These resources are now in
[...]

PR #359, opened 2023-07-26
add import for CSI volumes and job

lgfa29:
~No changelog since this is a new resource.~ 

Highjacking the PR to also add import support for jobs since I noticed that these are the last resources missing import.

Closes https://github.com/hashicorp/terraform-provider-nomad/issues/137

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ IMPROVEMENTS:
 * provider: add `skip_verify` configuration to skip TLS verification ([#319](https://github.com/hashicorp/terraform-provider-nomad/pull/319))
 * provider: update Go to 1.20.5 ([#334](https://github.com/hashicorp/terraform-provider-nomad/pull/334))
 * resource/nomad_acl_policy: add support for `job_acl` ([#314](https://github.com/hashicorp/terraform-provider-nomad/pull/314))
+* resource/nomad_csi_volume and resource/nomad_csi_volume_registration: add support to import existing volumes. ([#359](https://github.com/hashicorp/terraform-provider-nomad/pull/359)]
 * resource/nomad_job: add support to import existing jobs. ([#359](https://github.com/hashicorp/terraform-provider-nomad/pull/359)]
 * resource/nomad_volume and resource/nomad_external_volume: add timeouts for volume creation, registration, deregistration, and deletion ([#346](https://github.com/hashicorp/terraform-provider-nomad/pull/346))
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ IMPROVEMENTS:
 * provider: add `skip_verify` configuration to skip TLS verification ([#319](https://github.com/hashicorp/terraform-provider-nomad/pull/319))
 * provider: update Go to 1.20.5 ([#334](https://github.com/hashicorp/terraform-provider-nomad/pull/334))
 * resource/nomad_acl_policy: add support for `job_acl` ([#314](https://github.com/hashicorp/terraform-provider-nomad/pull/314))
+* resource/nomad_job: add support to import existing jobs. ([#359](https://github.com/hashicorp/terraform-provider-nomad/pull/359)]
```
tgross, line 21: And volumes!

tgross (commented):
Running the `terraform import` command seems to work fine for me here:

```
$ terraform import nomad_job.httpd httpd
nomad_job.httpd: Importing from ID "httpd"...
nomad_job.httpd: Import prepared!
  Prepared nomad_job for import
nomad_job.httpd: Refreshing state... [id=httpd]

Import successful!

The resources that were imported are shown above. These resources are now in
your Terraform state and will henceforth be managed by Terraform.
```

But using an [import](https://developer.hashicorp.com/terraform/language/import) does not, and I'm not sure if there's something extra we need to implement to make it work. Using the following Terraform config:

```hcl
provider "nomad" {
  address = "http://localhost:4646"
}

import {
  to = nomad_job.httpd
  id = "httpd"
}

resource "nomad_job" "httpd" {
}
```

I get the following error:

```
$ terraform plan
╷
│ Warning: Provider development overrides are in effect
│
│ The following provider development overrides are set in the CLI configuration:
│  - hashicorp/nomad in /home/tim/go/bin
│
│ The behavior may therefore not match any released version of the provider and applying
│ changes may cause the state to become incompatible with published releases.
╵
╷
│ Error: Missing required argument
│
│   on main.tf line 10, in resource "nomad_job" "httpd":
│   10: resource "nomad_job" "httpd" {
│
│ The argument "jobspec" is required, but no definition was found.
╵
```

lgfa29:
Ah yes, you need to add the `jobspec` attribute pointing to your jobfile. Something like this should work (assuming the job filename is correct and it lives in the same path as your `.tf` file):

```hcl
provider "nomad" {
  address = "http://localhost:4646"
}

import {
  to = nomad_job.httpd
  id = "httpd"
}

resource "nomad_job" "httpd" {
  jobspec = file("${path.module}/httpd.nomad.hcl")
}
```

tgross (approved):
LGTM!

tgross:
> Ah yes, you need to add the jobspec attribute pointing to your jobfile.

Oh... I guess that makes sense because the jobspec is external to the Terraform module, rather than being a set of attributes on the resource.

CHANGELOG.md
lgfa29: At first I thought we didn't need one, since the `nomad_csi_volume*` resources are new, but they actually replace the previous volume resource so it's probably a good idea to add a note for them as well 👍

lgfa29:
Yeah, after importing your config needs to match your state to prevent values from being lost in the next apply. There is experimental work on auto-generating configuration on import (https://developer.hashicorp.com/terraform/language/import/generating-configuration) but I doubt it would work for `nomad_job` since it's a weird resource with just a big string as input 😅

Issue #362, opened 2023-07-27
Empty retryable error received


blmhemu:
Same setup as https://github.com/hashicorp/terraform-provider-nomad/issues/349
Terraform resource is
```tf
resource "nomad_csi_volume" "alpine-caddy" {
  depends_on = [data.nomad_plugin.nfs]
  plugin_id  = data.nomad_plugin.nfs.plugin_id
  name       = "alpine-caddy"
  volume_id  = "alpine-caddy"
  capability {
    access_mode     = "multi-node-multi-writer"
    attachment_mode = "file-system"
  }
}

```
Initially I got
```
│ Error: invalid value 'capacity_min': strconv.ParseFloat: parsing "": invalid syntax
│ 
│   with nomad_csi_volume.alpine-caddy,
│   on main.tf line 34, in resource "nomad_csi_volume" "alpine-caddy":
│   34: resource "nomad_csi_volume" "alpine-caddy" {

```
even though that parameter is optional. (Root cause seems to be no empty checks at [this](https://github.com/hashicorp/terraform-provider-nomad/blame/main/nomad/resource_csi_volume.go#L331) and [this](https://github.com/hashicorp/terraform-provider-nomad/blame/main/nomad/resource_csi_volume.go#L336))

I have added capacity_min and capacity_max and then got the following issues.
```
│ Error: empty retryable error received. This is a bug with the Terraform provider and should be reported as a GitHub issue 
in the provider repository.
│ 
│   with nomad_csi_volume.alpine-caddy,
│   on main.tf line 34, in resource "nomad_csi_volume" "alpine-caddy":
│   34: resource "nomad_csi_volume" "alpine-caddy" {

```

lgfa29:
Thanks for the report @blmhemu.

The retryable error was fixed in https://github.com/hashicorp/terraform-provider-nomad/pull/354 but `capacity_min` and `capacity_max` should've been optional fields. This is fixed in #363.

blmhemu:
Thanks for the quick fix(es) ! Waiting for a new release :)

lgfa29:
v2.0.0-rc.1 has been released with the fix. I ma take a bit for the Terraform registry to pick up the new version.

PR #363, opened 2023-07-27
volumes: fix capacity parsing to allow optional values

lgfa29:
Allow `capacity_min` and `capacity_max` to actually be optional values.

Closes https://github.com/hashicorp/terraform-provider-nomad/issues/362


blmhemu (approved):

gulducat (approved):
to be sure, I confirmed these capacity values are optional both in the [CSI spec](https://github.com/container-storage-interface/spec/blob/b01039c563108173c6743aa1410ec11fde7c24fe/spec.md#createvolume) (CapacityRange) and the [nomad volume spec](https://developer.hashicorp.com/nomad/docs/other-specifications/volume#capacity_min) :+1:

lgfa29:
> to be sure, I confirmed these capacity values are optional both in the [CSI spec](https://github.com/container-storage-interface/spec/blob/b01039c563108173c6743aa1410ec11fde7c24fe/spec.md#createvolume) (CapacityRange) and the [nomad volume spec](https://developer.hashicorp.com/nomad/docs/other-specifications/volume#capacity_min) 👍

Thank you for double-checking. They were supposed to be optional 😬

PR #346, opened 2023-07-12
volumes: add timeout and retry on create and delete

lgfa29:
Add retry logic during volume create, register, deregister and delete.

Update deprecated resources since the deprecation has not happen yet, so keep them updated until v2.0.0 is out.

Closes https://github.com/hashicorp/terraform-provider-nomad/issues/335

PR #370, opened 2023-08-15
volume: revert retry logic on deprecated volume

lgfa29:
Revert retry logic added in #346 to avoid further issues with the existing volume resources (`nomad_external_volume` and `nomad_volume`). 

The new `nomad_csi_volume` and `nomad_csi_volume_registration` resources will have retries, but they require the use of the context CRUD methods (#371), which is a bigger change that we should not do for the deprecated resources.


jrasell (approved):
LGTM!

PR #371, opened 2023-08-15
volume: use context methods

### LINKED ISSUE #367: Ctrl + C interrupt ignored when waiting for volume deletion.
Consider the following case.
1. Setup CSI plugin.
2. In terraform, create Volumes.
3. Delete CSI controller / node => Plugin is unhealthy.
4. In terraform, remove the volume resource (from step 2)
5. Now terraform could not delete volume due to unavailable plugin and waits for it. 

Now you realize that the csi plugin is not active and try to interrupt terraform apply - it does not work, the retry still continues (till the time limit is reached ~ 5 min).

lgfa29:
Upddate `nomad_csi_volume` and `nomad_csi_volume_registration` to use CRUD methods with context and allow retries to be cancelled.

The read method was not updated because the signature clashes with the RetryContextFunc and there is not a nice way to transform a `diag.Diagnostics` to an `error`.

No changelog since these are unreleased resources.

Closes #367

### DIFF
--- a/nomad/resource_csi_volume.go
+++ b/nomad/resource_csi_volume.go
@@ -5,7 +5,7 @@ package nomad
 
 import (
 	"bytes"
-	"errors"
+	"context"
 	"fmt"
 	"hash/crc32"
 	"log"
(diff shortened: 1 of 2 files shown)


jrasell (approved):
LGTM!

PR #372, opened 2023-08-15
job: fix HCL2 enabled default value

### LINKED ISSUE #368: misleading message about hcl2 "enabled" going to `false`
### Terraform Version
$ terraform -v
Terraform v1.5.5
on linux_amd64
+ provider registry.terraform.io/hashicorp/aws v5.11.0
+ provider registry.terraform.io/hashicorp/nomad v2.0.0

### Nomad Version
1.6.1

### Provider Configuration

Provider version: 2.0.0-rc.1 (saved as file name 2.0.0 in the local cache inside `~/.terraform.d/...`

### Affected Resource(s)
- nomad_job

### Actual Behavior

When I started using the version 2.0.0-rc.1 (renamed to 2.0.0 on disk) ... I see the following ...

```bash
│ Starting with version 2.0.0 of the Nomad provider, jobs are parsed using HCL2 by default, so this field is no longer used and may be safely removed from your
│ configuration files. Set 'hcl1 = true' if you must use HCL1 job parsing.

```

... so I removed the line `enabled = true` inside the `hcl2` stanza.

Now when I run `terraform plan`, I see the following ...

```hcl
  ~ resource "nomad_job" "myjob" {
      + hcl1                    = false
        id                      = "myjob"
        name                    = "myjob"
      + read_allocation_ids     = false
        # (12 unchanged attributes hidden)

      ~ hcl2 {
          ~ enabled  = true -> false
            # (2 unchanged attributes hidden)
        }
    }

```

### Expected Behavior

As per the warning message, hcl2 is "enabled" by default, so if it is unspecified, it should be treated as `true`.

The `plan` output shows it going to `false`. Is this expected?

### Steps to Reproduce

using the v2.0.0-rc.1 provider version do "plan" and "apply" the job by explicitly setting `enabled = true` in the configuration file.
the remove the line `enabled = true` as suggested in the warning message.

### Important Factoids
provider version 2.0.0-rc.1 has been downloaded and placed inside local cache in `~/.terraform.d/plugins/registry.terraform.io/hashicorp/nomad/2.0.0/linux_amd64/terraform-provider-nomad_v2.0.0_x5`

lgfa29:
The `enabled` value inside the `hcl2` block is deprecated and not used anywhere in the code and HCL2 parsing is the new default, so the schema default value should reflect that to avoid incorrect plan results.

Closes #368

### DIFF
--- a/nomad/resource_job.go
+++ b/nomad/resource_job.go
@@ -98,7 +98,7 @@ func resourceJob() *schema.Resource {
 							Deprecated:  "Starting with version 2.0.0 of the Nomad provider, jobs are parsed using HCL2 by default, so this field is no longer used and may be safely removed from your configuration files. Set 'hcl1 = true' if you must use HCL1 job parsing.",
 							Type:        schema.TypeBool,
 							Optional:    true,
-							Default:     false,
+							Default:     true,
 						},
 						"allow_fs": {
 							Description: "If true, HCL2 file system functions will be enabled when parsing the `jobspec`.",


jrasell (approved):
LGTM!

PR #374, opened 2023-08-17
variable: fix 404 handling and remove deprecated methods

### LINKED ISSUE #373: Plan errors out when Nomad variable was deleted outside of Terraform
### Affected Resource(s)
- nomad_variable

### Expected Behavior
Variable should be detected as deleted outside of terraform and terraform should offer to recreate it

### Actual Behavior
Error: error checking for variable nomad/jobs/autoscaler@default: &errors.errorString{s:"variable not found"}
│
│   with nomad_variable.variable["autoscaler"],
│   on jobs.tf line 67, in resource "nomad_variable" "variable":
│   67: resource "nomad_variable" "variable" {

### Steps to Reproduce
1. Add `nomad_variable` resource in terraform
2. Run apply
3. Delete Nomad variable via UI/API
4. Run terraform plan

lgfa29:
Unlike other methods, `client.Variables().Read()` returns a special error when the path queried doesn't exist. I wasn't able to validate all code paths, so I kept the explicit `404` check in place just in case.

This resource was also created before the SDKv2 upgrade, so some of the methods used were deprecated. This PR also updates them to use the new methods.

No changelog since this resource has not been released yet.

Closes #373

### DIFF
--- a/nomad/resource_variable.go
+++ b/nomad/resource_variable.go
@@ -4,12 +4,15 @@
 package nomad
 
 import (
+	"errors"
 	"fmt"
 	"log"
 	"regexp"
 	"strings"
 
+	"github.com/hashicorp/go-cty/cty"
 	"github.com/hashicorp/nomad/api"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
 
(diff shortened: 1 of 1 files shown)


nomad/resource_variable.go
```diff
@@ -168,24 +176,37 @@ func pathValidation() schema.SchemaValidateFunc {
...
 		}
 
 		if len(parts) == 1 {
-			errs = append(errs, fmt.Errorf("path of 'nomad' is a reserved top-level directory path"))
+			diags = append(diags, diag.Diagnostic{
+				Severity:      diag.Error,
+				Summary:       "path of 'nomad' is a reserved top-level directory path",
+				AttributePath: k,
+			})
+			return diags
```
lgfa29, line 188: Without this early return the `switch` below panics.

jrasell (approved):

PR #382, opened 2023-09-19
Allow increasing the size of CSI volumes

gulducat:
Nomad is enabling CSI volume expansion per hashicorp/nomad#10324 -- slated for the next `1.6.x` release (looks like `1.6.3` specifically)

To that end here, `capacity_min` and `capacity_max` changes will no longer force replacement of `nomad_csi_volume`s, `nomad_csi_volume_registration`s can take those values too, and updating them appropriately can trigger a volume expansion for CSI plugins that support it.

On older versions of Nomad, this _should_ produce idempotent Create or Register API calls (for `nomad_csi_volume` and `nomad_csi_volume_registration` respectively), and if this TF provider detects that expansion has not occurred (by comparing against actual Capacity as returned by Nomad API), it will issue a warning to the user.

I did not update `nomad_external_volume` as it's marked as deprecated, and honestly I considered excluding `nomad_csi_volume_registration` too, because it didn't take capacity min/max to begin with.  But!  It shares chunks of code with `noamd_csi_volume`, and I did update the Register Nomad API to behave the ~same as Create so registration came along too.

There's quite a bit of unit test coverage of the new supporting functions, but very minimal change to the acceptance test.  For example, from what I can tell there is nothing testing Update operations at all, and I have not really improved that situation.  I can take a swing at it though if anyone would like to push me to.


nomad/resource_csi_volume.go
```diff
@@ -330,30 +346,8 @@ func resourceCSIVolumeCreate(ctx context.Context, d *schema.ResourceData, meta i
...
-		var err error
-		capacityMax, err = humanize.ParseBytes(capacityMaxStr)
-		if err != nil {
-			parsingDiags = append(parsingDiags, diag.Diagnostic{
-				Severity: diag.Error,
-				Summary:  fmt.Sprintf("invalid value 'capacity_max': %v", err),
-			})
-		}
-	}
+	capacityMin, capacityMax, capacityDiags := parseCapacity(d)
```
Juanadelacuesta, line 349: nice refactor!

nomad/resource_csi_volume_test.go
```diff
@@ -227,3 +235,202 @@ resource "nomad_csi_volume" "test" {
...
+func TestCapacityStateFunc(t *testing.T) {
+	cases := []struct {
+		in, out string
+	}{
+		{"5", "5 B"},
+		{"1kb", "1000 B"},
+		{"10kib", "10 KiB"},
+		{"5.5 Gib", "5.5 GiB"},
+		{"5GB", "4.7 GiB"},
+		{"ugh", "ugh"}, // validation happens elsewhere
```
Juanadelacuesta, line 278: lovely test values jejeje

Juanadelacuesta: no test for empty values?

Juanadelacuesta (approved):
Gave up on the acceptance tests? The rest looks good

nomad/resource_csi_volume_test.go
gulducat: oh sure fair point, here go: cc1511a

gulducat:
I didn't really start adding steps to acceptance tests, but your question @Juanadelacuesta was enough for me to go ahead and poke at it.

The current challenge I have with it is that the publicly-available version of Nomad doesn't support volume expansion, and the `hostpath` CSI plugin that we use here for the tests throws an error about the size changing on an existing volume:

>│ Error: error creating CSI volume: Unexpected response code: 500 (1 error occurred:
>│       * controller create volume: CSI.ControllerCreateVolume: volume "mysql_volume" already exists but is incompatible with these parameters: rpc error: code = AlreadyExists desc = Volume with the same name: mysql_volume but with different size already exist)

and it seems like the test framework -- specifically [this `Config` field](https://github.com/hashicorp/terraform-provider-nomad/blob/a395311/nomad/resource_csi_volume_test.go#L33) -- wants applying the config not to error.

All that to say, I think updating the acceptance test to confirm changes will make more sense after volume expansion in Nomad is released (and so available to pull in here).

Issue #274, opened 2022-07-07
TFC overrides the default region("global") by "tf-cloud-iso-prod-use1"


lens0021:
### Terraform Version
`Terraform v1.2.4`

### Nomad Version

`Nomad v1.3.1 (2b054e38e91af964d1235faa98c286ca3f527e56)`

### Provider Configuration

```hcl
provider "nomad" {
  address   = var.nomad_address
}

resource "nomad_job" "memcached" {
  jobspec = file("memcached.nomad")
  detach  = false
}
```

`memcached.nomad`:
```hcl
job "memcached" {
  datacenters = ["dc1"]

  group "memcached" {
    task "memcached" {
      driver = "docker"

      config {
        image = "memcached:1.6.15-alpine"
      }
    }
  }
}
```

([See the full code here](https://gitlab.com/lens0021/terraform-nomad-provider-workspace-sample))

### Debug Output
> **Error: error applying jobspec: Unexpected response code: 500 (No path to region)**
with nomad_job.memcached
on base.tf line 29, in resource "nomad_job" "Memcached":
> ```
> resource "nomad_job" "memcached" {
> ```

### Panic Output

> 2022-07-07T10:46:05.769+0900 [WARN]  nomad.rpc: no path found to region: region=tf-cloud-iso-prod-use1
> 2022-07-07T10:46:05.770+0900 [ERROR] http: request failed: method=GET path="/v1/job/memcached?namespace=default&region=tf-cloud-iso-prod-use1" error="No path to region" code=500

### Expected Behavior
I didn't set the region, so the default value `"global"` should be used.

### Actual Behavior
The region seems to be set to `"tf-cloud-iso-prod-use1"`.

### Steps to Reproduce

1. Setup a workspace with the Remote execution mode.
2. Start new run on TFC.
3. Apply.

And as a workaround:

1. Adding `region = "global"` makes the problem disappear.

### References

GH-267 is very similar to this. The problem has shown up without code changes since someday.

oinkbark:
I had basically the same issue of the default "global" region being overridden, but it was simply not being set to anything.
This was the error that was producing for all of my jobs:
```
Error: error checking for job: Unexpected response code: 500 (No path to region)
```
Provider version: 1.4.17 (Previously 1.4.15)
Nomad Version: 1.3.2 (Previously 1.0.4)

[...]

PR #386, opened 2023-10-06
resource/job: set namespace on register

FourLeafTec:
When register jobs by Http api need to set the job namespace in the request.

coder/coder#10074

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ IMPROVEMENTS:
 BUG FIXES:
 * resource/nomad_acl_policy: fixed a bug where the namespace would be incorrectly calculated from a job identity ([#396](https://github.com/hashicorp/terraform-provider-nomad/pull/396))
 * resource/nomad_csi_volume_registration: fixed a bug that cause an import operation to not load all of the volume attributes ([#402](https://github.com/hashicorp/terraform-provider-nomad/pull/402))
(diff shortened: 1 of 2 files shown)


lgfa29:
Hi @FourLeafTec 👋 

Thank you for the PR, and apologies for the delay on getting this reviewed. I think the changes look good, but I'm curious to understand how you reached this problem.

The Nomad API should use the namespace from the job if none is set in the request:
https://github.com/hashicorp/nomad/blob/a6e164673ed5f37f010eafb8ff0adc92de195946/command/agent/job_endpoint.go#L1002-L1019

Since the provider doesn't set any `WriteOptions` I'm curious how you ended up with the request reported in https://github.com/coder/coder/issues/10074#issuecomment-1751072918, with the `?namespace=default` query param.

```
[DEBUG] http: request complete: method=PUT path="/v1/jobs?namespace=default&region=global"
```

I have not use `coder` before (looks like a cool project!) but does it not use Terraform directly?

lgfa29:
Ohh I see, I think it's the same problem as https://github.com/hashicorp/terraform-provider-nomad/issues/274. 

If `coder` is running the `terraform` CLI as a Nomad job, then it may be reading the `NOMAD_NAMESPACE` env var from that allocation.

We added the `ignore_env_vars` to fix this problem, so you can update your provider config to something like this:

```hcl
provider "nomad" {
  ignore_env_vars = {
    "NOMAD_NAMESPACE" = true
    "NOMAD_REGION"    = true
  }
}
```

But setting the correct namespace sounds like a good thing to do regardless. Just watch out for the `NOMAD_REGION` env var 😅

lgfa29 (approved):
Pushed a commit to also set the namespace on `PlanOpts` and another one to add a changelog entry.

PR #389, opened 2023-10-13
job: ensure that job is running

rodrigol-chan:
Refs #1. It's sometimes handy to use the UI to stop jobs e.g. during a maintenance event. In this specific case, the provider should check if the job is no longer running and resubmit it. I'm open to suggestions on how to approach this better.

### DIFF
--- a/nomad/resource_job.go
+++ b/nomad/resource_job.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"log"
 	"reflect"
+	"slices"
 	"sort"
 	"strconv"
 	"strings"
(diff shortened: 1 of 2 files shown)


rodrigol-chan:
@lgfa29 Sorry for the ping, but is there interest in this? We would really like to solve the "ensure job is running" problem.

lgfa29:
Hi @rodrigol-chan 👋 

Apologies for the delay in review the PR. We're wrapping work for Nomad 1.7, which has kept us busy for the past weeks. I will try to review this some time next week.

Thanks for raising it!

lgfa29 (approved):
Thanks for the PR @rodrigol-chan, and apologies again for the delay here. 

I think the changes look good, but I do worry about this breaking backwards compatibility in case someone is relying on the existing behaviour.

I added a new flag called `rerun_if_dead` to allow for this new behaviour to be opt-in but I couldn't push the change to your branch 😅 

Since this PR is atomic enough I will merge it as-is and then open a follow-up PR for the new attribute.

Thank you again for the contribution!

Issue #390, opened 2023-10-20
Import of nomad_csi_volume_registration lack a some piece of information


luhhujbb:
### Terraform Version
1.6.2

### Nomad Version
1.6.2

### Provider Configuration
Which values are you setting in the provider configuration? 

```hcl
terraform {
  required_providers {
    nomad = {
      source  = "hashicorp/nomad"
      version = ">= 2.0.0"
    }
  }
}

```

### Environment Variables
N/A

### Affected Resource(s)
- nomad_csi_volume_registration

### Terraform Configuration Files
```hcl
data "nomad_plugin" "gcepd" {
  plugin_id        = "gcepd"
  wait_for_healthy = true
}

resource "google_compute_disk" "mongo_volume" {
  name = "mongo-${var.mongo_name}-${var.env}"
  type = var.mongo_volume_type
  zone = "${var.region}-${var.zones[0]}"
  labels = {
    env = var.env
  }
  physical_block_size_bytes = 4096
  size                      = var.mongo_volume_size
}

resource "nomad_csi_volume_registration" "mongo_volume" {
  depends_on  = [data.nomad_plugin.gcepd]
  plugin_id   = "gcepd"
  volume_id   = "mongo-${var.mongo_name}"
  name        = "mongo-${var.mongo_name}"
  external_id = google_compute_disk.mongo_volume.id

  topology_request {
    required {
      topology {
        segments = {
          "topology.gke.io/zone" = google_compute_disk.mongo_volume.zone
        }
      }
    }
  }

  capability {
    access_mode     = "single-node-writer"
    attachment_mode = "file-system"
  }

  mount_options {
    fs_type = "xfs"
  }
}

```

### Expected Behavior
Import should populate all necessary fields

### Actual Behavior
When migrating from nomad_volume to nomad_csi_volume_registration, we need to import nomad_csi_volume_registration, and the following properties are not well imported:
* namespace
* plugin_id
* volume_id
* external_id
* capability
* mount_options
* deregister_on_destroy

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. `terraform import nomad_csi_volume_registration.mongo_volume`
2. `terraform plan`
3. It displays the following changes :
```
-/+ resource "nomad_csi_volume_registration" "mongo_volume" {
      ~ controller_required     = true -> (known after apply)
      ~ controllers_expected    = 1 -> (known after apply)
      ~ controllers_healthy     = 1 -> (known after apply)
[...]

PR #401, opened 2023-12-12
Set fields when importing csi volume registrations

gastrodon:
Create resourceCSIVolumeRegistrationImport, and set some of the fields that terraform otherwise wants to re-create the registration over

Related to #390


gastrodon:
doing this on a different branch..

PR #402, opened 2023-12-12
Set fields when importing csi volume registrations

gastrodon:
Create `resourceCSIVolumeRegistrationImport`, and set some of the fields that terraform otherwise wants to re-create the registration over

Related to #390


lgfa29 (approved):
Thank you for the PR @gastrodon! 

Your analysis was spot on, but the fix was actually a little simpler. The missing attributes were actually missing from `resourceCSIVolumeRead`, which is what the passthrough importer uses.

I pushed a commit with this approach and also added a changelog entry.

Thanks again for the contribution!

lgfa29:
Oops...I forgot `resourceCSIVolumeRead` is shared between volume and volume registration so we can't set the `external_id` there 😅 

I will push a commit to fix it tomorrow.

Issue #394, opened 2023-11-28
acl_auth_method: add token name format parameter


jrasell:
Nomad PR https://github.com/hashicorp/nomad/pull/19135 added support for an optional token name format parameter within an ACL authentication method. The Nomad Terraform provider should be updated to support this.

the-nando:
hey @jrasell :) did you get a chance to look at this? I can put together an MR otherwise

jrasell:
Hi @the-nando; I have not got around to this. If you have the ability to raise a PR with this addition, that would be amazing. If you have any questions, feel free to ping me about it.

the-nando:
@jrasell I've tested it on a local cluster and I get the expected result:
```
resource "nomad_acl_auth_method" "test" {
  name           	  = "foo"
  type           	  = "OIDC"
  token_locality 	  = "global"
  token_name_format = "$${auth_method_type}-$${auth_method_name}-$${value.user}"
  max_token_ttl  	  = "10m0s"
  default        	  = "false"

  config {
    oidc_discovery_url    = "https://uk.auth0.com/"
    oidc_client_id        = "someclientid"
    oidc_client_secret    = "someclientsecret-t"
    oidc_scopes           = ["email"]
    bound_audiences       = ["someclientid"]
    discovery_ca_pem      = ["secretpemcert"]
    signing_algs          = ["rsa256", "hs256"]
    allowed_redirect_uris = [
      "http://localhost:4649/oidc/callback",
    ]
    claim_mappings = {
      "http://nomad.internal/name": "name"
    }
    list_claim_mappings = {
      "http://nomad.internal/roles": "roles"
    }
  }
}
```
```
~ curl -s -N --cacert ${NOMAD_CACERT} --cert ${NOMAD_CLIENT_CERT} --key ${NOMAD_CLIENT_KEY} -H "X-Nomad-Token: ${NOMAD_TOKEN}" ${NOMAD_ADDR}/v1/acl/auth-method/foo | jq .TokenNameFormat                                  
"${auth_method_type}-${auth_method_name}-${value.user}"
~
```

The acceptance tests pass as well when run locally but fail when run in [GH](https://github.com/the-nando/terraform-provider-nomad/actions/runs/7205601540/job/19628882718):
```
    resource_acl_auth_method_test.go:26: Step 1/2 error: Check failed: expected token name format to be "${auth_method_type}-${auth_method_name}-${value.user}", is "" in API
[...]

PR #403, opened 2023-12-13
Bump to 1.7.1 and add token_name_format to resource_acl_auth_method

the-nando:
Fixes https://github.com/hashicorp/terraform-provider-nomad/issues/394


lgfa29 (approved):
Thank you for the great PR @the-nando!

I had updated Nomad to 1.7.2 in #404 so I rebased and force-pushed to your branch to pick that up. Releases have been rolling out quickly 😬

IamTheFij:
Noting that this introduces a bug. This field is optional without a default value, yet the server provides a default value when requesting the resource. This results in Terraform always thinking there is a change and “removing” the value. 

For example:

```

  # module.core.nomad_acl_auth_method.nomad_authelia will be updated in-place
  ~ resource "nomad_acl_auth_method" "nomad_authelia" {
        id                = "authelia"
        name              = "authelia"
      - token_name_format = "${auth_method_type}-${auth_method_name}" -> null
        # (4 unchanged attributes hidden)

        # (1 unchanged block hidden)
    }

module.core.nomad_acl_auth_method.nomad_authelia: Modifying... [id=authelia]
module.core.nomad_acl_auth_method.nomad_authelia: Modifications complete after 0s [id=authelia]

Apply complete! Resources: 0 added, 1 changed, 0 destroyed.
```

I’ll report an issue as well when I have a bit more time, but I thought it may help to provide this along with the context of the original change as well.

lgfa29:
Thanks for the report @IamTheFij!

Sorry I missed this earlier, GitHub notifications can be a little weird, so opening an issue and linking to the PR is usually better way to get our attention 😅 

I've opened #442 to fix the problem.

PR #407, opened 2023-12-15
job: add rerun_if_dead argument 

lgfa29:
Follow-up work from https://github.com/hashicorp/terraform-provider-nomad/pull/389#pullrequestreview-1785043456 to introduce a new argument called `rerun_if_dead` to allow users to opt-in to the new behaviour.


gulducat (approved):
I wish we could know if changing the default behavior to this would actually be disruptive to anyone, but short of that, this looks like a nice new option!

lgfa29:
We're trying to be more strict with semver on the provider, so I think a change in the default behavour would need to be considered a breaking change.

We can revisit this base on user feedback.

PR #375, opened 2023-08-18
import: fall-back to namespace from provider

Chive:
# Summary

Resolves an issue with importing resources from non-default namespaces. It was impossible to import resources from namespaces other than default, because opts.Namespace would always be `""`, discarding any environment variable set for `NOMAD_NAMESPACE`.

# Reproducing the issue

## Prerequisites
- Running Nomad agent
- Created Nomad namespace (i.e. `foo`)

### Files
<details>
<summary>Expand</summary>

#### `main.tf`
```terraform
terraform {
  required_providers {
    nomad = {
      source = "hashicorp/nomad"
      version = "2.0.0-rc.2"
    }
  }
}

resource "nomad_job" "redis" {
  jobspec               = file("redis.nomad")
  detach                = false
  deregister_on_destroy = false
}
```

#### `redis.nomad`

```hcl
job "redis" {
  type  = "service"
  namespace = "foo"
  region = "region"

  group "redis" {
    count = 1
    
    task "redis" {
      driver = "docker"

      config {
        image = "redis"
      }
    }
  }
}
```
</details>

### Environment
<details>
<summary>Expand</summary>

#### Env Vars
```sh
$ env | grep NOMAD_
NOMAD_TOKEN=token
NOMAD_ADDR=http://nomad:4646
NOMAD_REGION=region
NOMAD_NAMESPACE=foo
```

#### Nomad job status
```
$ nomad job status -short redis
ID            = redis
Name          = redis
Type          = service
Namespace     = foo
Status        = running
[...]
```
</details>

## Running `terraform import` on version `2.0.0-rc.2`
<details>
<summary>Expand</summary>

```sh
$ terraform import nomad_job.redis redis
2023-08-18T11:15:19.937+0200 [INFO]  Terraform version: 1.5.5
2023-08-18T11:15:19.937+0200 [DEBUG] using github.com/hashicorp/go-tfe v1.26.0
2023-08-18T11:15:19.937+0200 [DEBUG] using github.com/hashicorp/hcl/v2 v2.16.2
2023-08-18T11:15:19.937+0200 [DEBUG] using github.com/hashicorp/terraform-svchost v0.1.0
2023-08-18T11:15:19.937+0200 [DEBUG] using github.com/zclconf/go-cty v1.12.2
2023-08-18T11:15:19.937+0200 [INFO]  Go runtime version: go1.20.7
2023-08-18T11:15:19.937+0200 [INFO]  CLI args: []string{"terraform", "import", "nomad_job.redis", "redis"}
[...]

PR #408, opened 2023-12-16
job: import jobs from non-default namespace

lgfa29:
`nomad_job`, `nomad_csi_volume`, and `nomad_csi_volume_registration` do not have the namespace as part of their state ID. This makes it impossible to import them if they are registered in a non-default namespace since only the ID is provided to the import function.

This commit adds a new helper that imports using an ID with the format `<id>@<namespace>`. Since `@` is not a valid character for namespaces, the last `@` character (if found) represents the separator.

Implements the approach suggested in https://github.com/hashicorp/terraform-provider-nomad/pull/375#issuecomment-1686797489.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ IMPROVEMENTS:
 * resource/nomad_csi_volume: changes to `capacity_min` or `capacity_max` may now expand the volume instead of forcing replacement,
   on Nomad version 1.6.3 or later, if the CSI plugin supports it ([#382](https://github.com/hashicorp/terraform-provider-nomad/pull/382))
 * resource/nomad_job: Add `rerun_if_dead` attribute to allow forcing a job to run again if it's marked as `dead`. ([#407](https://github.com/hashicorp/terraform-provider-nomad/pull/407))
+* resource/nomad_job: add support for importing jobs from non-default namespace ([#408](https://github.com/hashicorp/terraform-provider-nomad/pull/408))
 
 BUG FIXES:
 * resource/nomad_acl_policy: fixed a bug where the namespace would be incorrectly calculated from a job identity ([#396](https://github.com/hashicorp/terraform-provider-nomad/pull/396))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,9 +5,11 @@ IMPROVEMENTS:
 * provider: update Go to 1.21.5 ([#399](https://github.com/hashicorp/terraform-provider-nomad/pull/399))
 * resource/resource_acl_auth_method: add support for `token_name_format` ([#403](https://github.com/hashicorp/terraform-provider-nomad/pull/403))
 * resource/nomad_csi_volume: changes to `capacity_min` or `capacity_max` may now expand the volume instead of forcing replacement,
+* resource/nomad_csi_volume: update import key to be `<volume id>@<namespace>` to allow importing volumes from namespaces other than `default` ([#408](https://github.com/hashicorp/terraform-provider-nomad/pull/408))
+* resource/nomad_csi_volume_registration: update import key to be `<volume id>@<namespace>` to allow importing volume registrations from namespaces other than `default` ([#408](https://github.com/hashicorp/terraform-provider-nomad/pull/408))
   on Nomad version 1.6.3 or later, if the CSI plugin supports it ([#382](https://github.com/hashicorp/terraform-provider-nomad/pull/382))
 * resource/nomad_job: Add `rerun_if_dead` attribute to allow forcing a job to run again if it's marked as `dead`. ([#407](https://github.com/hashicorp/terraform-provider-nomad/pull/407))
-* resource/nomad_job: add support for importing jobs from non-default namespace ([#408](https://github.com/hashicorp/terraform-provider-nomad/pull/408))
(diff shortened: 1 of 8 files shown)


lgfa29:
Tests seems to have started failing in https://github.com/hashicorp/terraform-provider-nomad/commit/a81b5f767126c14817494cd57279bf31efab12aa. I will investigate in a follow-up PR.

lgfa29:
Oh, I think https://github.com/hashicorp/terraform-provider-nomad/pull/407 fixes the tests. We'll see once all PRs are merged 😅

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ IMPROVEMENTS:
 * resource/nomad_csi_volume: changes to `capacity_min` or `capacity_max` may now expand the volume instead of forcing replacement,
   on Nomad version 1.6.3 or later, if the CSI plugin supports it ([#382](https://github.com/hashicorp/terraform-provider-nomad/pull/382))
 * resource/acl_auth_method: add support for `token_name_format` ([#403](https://github.com/hashicorp/terraform-provider-nomad/pull/403))
+* resource/nomad_job: add support for importing jobs from non-default namespace ([#408](https://github.com/hashicorp/terraform-provider-nomad/pull/408))
```
gulducat, line 9: This is a little misleading -- to me it implies no changes if you always use the default namespace, but we're actually making it _mandatory_ to include the namespace when importing these resources, even when it's `@default` (due to the challenging fact that job IDs can contain just about any character, including `@`).  Nice to lean positive with "add support" but I might also mention the new requirement.

Also, is it worth calling CSI volumes here too? I imagine they're much less often used, so maybe ok to leave out of the CL.

nomad/helper/import.go
```diff
@@ -0,0 +1,42 @@
...
+var (
+	missingNamespaceImportErr = errors.New("missing namespace, the import ID should follow the pattern <id>@<namespace>")
+	missingIDImportErr        = errors.New("missing resource ID, the import ID should follow the pattern <id>@<namespace>")
+)
+
+// NamespacedImporterContext imports a namespaced resource that doesn't have
+// its namespace as part of the Terraform resource ID.
+func NamespacedImporterContext(_ context.Context, d *schema.ResourceData, meta any) ([]*schema.ResourceData, error) {
+	namespacedID := d.Id()
+	sepIdx := strings.LastIndex(d.Id(), "@")
```
gulducat, line 23: super nit: you defined `namespacedID` just now but then didn't use it :P

```suggestion
	sepIdx := strings.LastIndex(namespacedID, "@")
```

website/docs/r/csi_volume.html.markdown
```diff
@@ -137,5 +137,22 @@ configuration options.
 - `create` `(string: "10m")` - Timeout when creating or updating a new CSI volume.
 - `delete` `(string: "10m")` - Timeout when deleting a CSI volume.
 
+## Importing CSI Volumes
+
+CSI volumes are imported using the pattern `<volume ID>@<namespace>` .
```
gulducat, line 142: I think these might benefit from some info on how to get the namespace, or short of that, saying ~"If you're not sure, it's probably the default, which is `@default`" and maybe using `@default` in the examples.

nomad/helper/import.go
```diff
@@ -0,0 +1,42 @@
...
+	if len(ns) == 0 {
+		return nil, missingNamespaceImportErr
+	}
+
+	id := namespacedID[:sepIdx]
+	if len(id) == 0 {
+		return nil, missingIDImportErr
+	}
+
+	d.SetId(id)
```
gulducat, line 38: I like that this retains the same non-namespaced ID in state, so no state migration necessary :+1:

gulducat (approved):
some food for thought, but LGTM!

CHANGELOG.md
lgfa29: I has always been possible to import resources from the `default` namespace, so the change is that non-default namespaces are now supported. But I see what you mean. Updated to:

> * resource/nomad_job: update import key to be `<job id>@<namespace>` to allow importing jobs from namespaces other than `default` ([#408](https://github.com/hashicorp/terraform-provider-nomad/pull/408))


Also added similar entries to volumes. I just forgot about them 😅

website/docs/r/csi_volume.html.markdown
lgfa29: Hum...not sure about this. If you're importing a resource you must know its namespace. If you don't, you really shouldn't be importing it 😬 

You will also need to set the namespace in the TF config, so you _really_ need to know it if you're trying to import it.

gulducat: For jobs, the namespace can be excluded from the jobspec, and it isn't an argument in the TF resource.  For CSI volumes, the namespace is a TF resource argument, but can also be left out and defaults to "default"

So our just-getting-started type folks, or even experienced users who just never needed namespaces for some reason, could totally get away with never thinking about namespaces at all -- the default "just works" -- so I'm thinking this might be a little learning opportunity where they could use just a bit of extra help.

lgfa29: But importing a resource in Terraform is not targeted at just-getting-started folks, it's more like a day 10321 type of thing 😅

PR #410, opened 2023-12-18
volume: add computed attribute `external_id`

lgfa29:
CSI volumes have an external ID regardless on how they were added to Nomad, so add a computed attribute `external_id` to `nomad_csi_volume`.

This makes the schema of `nomad_csi_volume` and
`nomad_csi_volume_registration` compatible so that the read method can be shared between then again.

Alternate (and arguably correct 😅) implementation of https://github.com/hashicorp/terraform-provider-nomad/pull/402

### DIFF
--- a/nomad/resource_csi_volume.go
+++ b/nomad/resource_csi_volume.go
@@ -9,8 +9,10 @@ import (
 	"fmt"
 	"hash/crc32"
 	"log"
+	"strings"
 	"time"
 
+	"github.com/dustin/go-humanize"
 	"github.com/hashicorp/nomad/api"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/retry"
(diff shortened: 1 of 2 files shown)


nomad/resource_csi_volume.go
```diff
@@ -336,13 +335,77 @@ func resourceCSIVolume() *schema.Resource {
...
+		}
+
+		return fmt.Errorf("error checking for CSI volume: %s", err)
+	}
+	log.Printf("[DEBUG] found CSI volume %q in namespace %q", volume.Name, volume.Namespace)
+
+	d.Set("name", volume.Name)
+	d.Set("namespace", volume.Namespace)
+	d.Set("volume_id", volume.ID)
+	d.Set("external_id", volume.ExternalID)
```
lgfa29, line 379: Moved the read logic to this resource since it seemed to make more sense name-wise. This was the only change.

gulducat (approved):
LGTM!

PR #412, opened 2023-12-19
job: prevent partial update on error

### LINKED ISSUE #385: nomad_job : provider updates state on failure
### Nomad Version
`1.6.2+ent`

### Provider Configuration

```hcl
provider "nomad" {
  alias   = "ap-northeast-1"
  address = "https://<addr>"
  region  = "ap-northeast-1"
  ignore_env_vars = {
    NOMAD_NAMESPACE = true
    NOMAD_REGION    = true
  }
}
```

### Affected Resource(s)
- nomad_job

```hcl
resource "nomad_job" "test-job" {
  jobspec = file("${path.module}/test-job.hcl")
}
```

```hcl
job "test-job" {
  datacenters = ["ap-northeast-1a"]

  namespace = "system"

  group "test" {

    task "test" {
      driver = "docker"

      config {
        image = "alpine:3.18"
        args  = ["tail", "-f", "/dev/null"]
      }

      resources {
        cpu    = 100
        memory = 64
      }
    }
  }
}
```

After applying the job with a Nomad token with enough permissions, I edit the job's spec and re-run Terraform with a Nomad token which lacks permissions:

```
module.test.nomad_job.test-job: Refreshing state... [id=test-job]

Terraform used the selected providers to generate the following execution plan. Resource actions are indicated with the following symbols:
  ~ update in-place

  # module.test.nomad_job.test-job will be updated in-place
  ~ resource "nomad_job" "test-job" {
      ~ allocation_ids          = [] -> (known after apply)
      ~ datacenters             = [
          - "ap-northeast-1b",
            # (1 unchanged element hidden)
        ]
        id                      = "test-job"
      ~ jobspec                 = <<-EOT
            job "test-job" {
          -   datacenters = ["ap-northeast-1a", "ap-northeast-1b"]
          +   datacenters = ["ap-northeast-1a"]

              namespace = "system"

              group "test" {

                task "test" {
                  driver = "docker"

                  config {
                    image = "alpine:3.18"
                    args  = ["tail", "-f", "/dev/null"]
                  }

                  resources {
                    cpu    = 100
                    memory = 64
                  }
                }
              }
            }
        EOT
      ~ modify_index            = "1172901" -> (known after apply)
        name                    = "test-job"
      ~ region                  = "ap-northeast-1" -> (known after apply)
      ~ type                    = "service" -> (known after apply)
        # (8 unchanged attributes hidden)
    }

Do you want to perform these actions?
  Terraform will perform the actions described above.
  Only 'yes' will be accepted to approve.

  Enter a value: yes

module.test.nomad_job.test-job: Modifying... [id=test-job]
╷
│ Error: error applying jobspec: Unexpected response code: 403 (Permission denied)
│
│   with module.test.nomad_job.test-job,
│   on ../../../modules/test/main.tf line 78, in resource "nomad_job" "test-job":
│   78: resource "nomad_job" "test-job" {
```
A subsequent plan runs clean, even though the job itself is still running with the old spec and the state is updated:
```bash
~ NOMAD_REGION=ap-northeast-1 nomad job inspect -namespace='system' test-job | jq -r .Job.Datacenters
[
  "ap-northeast-1a",
  "ap-northeast-1b"
]
~  terraform state show module.test.nomad_job.test-job
# module.test.nomad_job.test-job:
resource "nomad_job" "test-job" {
    datacenters             = [
        "ap-northeast-1a",
    ]
    deregister_on_destroy   = true
    deregister_on_id_change = true
    detach                  = true
    hcl1                    = false
    id                      = "test-job"
    jobspec                 = <<-EOT
        job "test-job" {
          datacenters = ["ap-northeast-1a"]

          namespace = "system"

          group "test" {

            task "test" {
              driver = "docker"

              config {
                image = "alpine:3.18"
                args  = ["tail", "-f", "/dev/null"]
              }

              resources {
                cpu    = 100
                memory = 64
              }
            }
          }
        }
    EOT
    modify_index            = "1172901"
    name                    = "test-job"
    namespace               = "system"
    policy_override         = true
    read_allocation_ids     = false
    region                  = "ap-northeast-1"
    task_groups             = [
        {
            count   = 1
            meta    = {}
            name    = "test"
            task    = [
                {
                    driver        = "docker"
                    meta          = {}
                    name          = "test"
                    volume_mounts = []
                },
            ]
            volumes = []
        },
    ]
    type                    = "service"
}
~
```

### Expected Behavior
The state isn't updated.

### Actual Behavior
The state is updated and diverges from reality.

lgfa29:
The default behaviour of the Terraform SDK is to copy the plan result into state which could result in partial state updates, where Terraform state is updated, but the actual resource state is not, in case of an error during the apply.

This is normally not an issue because resources are expected to undo these changes on state refresh. Any partial update is reconciled with the actual resource state.

But due to #1, the `nomad_job` resource is not able to properly reconcile on refresh, causing the partial update to prevent further applies unless the configuration is also changed.

This commit uses the `d.Partial()` method to signal to Terraform that any state changes should be rolledback in case of an error.

Closes #385

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ BUG FIXES:
 * resource/nomad_acl_policy: fixed a bug where the namespace would be incorrectly calculated from a job identity ([#396](https://github.com/hashicorp/terraform-provider-nomad/pull/396))
 * resource/nomad_csi_volume_registration: fixed a bug that cause an import operation to not load all of the volume attributes ([#402](https://github.com/hashicorp/terraform-provider-nomad/pull/402))
 * resource/nomad_job: fixed a bug that could cause jobs to be registered in the incorrect namespace if the `NOMAD_NAMESPACE` environment variable is set ([#386](https://github.com/hashicorp/terraform-provider-nomad/pull/386))
+* resource/nomad_job: fixed a bug that caused state changes even in case of errors during apply ([#412](https://github.com/hashicorp/terraform-provider-nomad/pull/412))
 * resource/nomad_volume: fixed a bug that cause an import operation to not load all of the volume attributes ([#402](https://github.com/hashicorp/terraform-provider-nomad/pull/402))
 
 ## 2.0.0 (August 28th, 2023)
--- a/nomad/resource_job.go
+++ b/nomad/resource_job.go
@@ -362,6 +362,7 @@ func resourceJobRegister(d *schema.ResourceData, meta interface{}) error {
 	timeout := d.Timeout(schema.TimeoutCreate)
 	if !d.IsNewResource() {
 		timeout = d.Timeout(schema.TimeoutUpdate)
+		d.Partial(true)
 	}
 
 	providerConfig := meta.(ProviderConfig)
@@ -413,6 +414,10 @@ func resourceJobRegister(d *schema.ResourceData, meta interface{}) error {
 	})
 	if err != nil {
 		return fmt.Errorf("error applying jobspec: %s", err)
+	}
+
+	if !d.IsNewResource() {
+		d.Partial(false)
 	}
 
 	log.Printf("[DEBUG] job '%s' registered in namespace '%s'", *job.ID, *job.Namespace)


CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ BUG FIXES:
 * resource/nomad_acl_policy: fixed a bug where the namespace would be incorrectly calculated from a job identity ([#396](https://github.com/hashicorp/terraform-provider-nomad/pull/396))
 * resource/nomad_csi_volume_registration: fixed a bug that cause an import operation to not load all of the volume attributes ([#402](https://github.com/hashicorp/terraform-provider-nomad/pull/402))
 * resource/nomad_job: fixed a bug that could cause jobs to be registered in the incorrect namespace if the `NOMAD_NAMESPACE` environment variable is set ([#386](https://github.com/hashicorp/terraform-provider-nomad/pull/386))
+* resource/nomad_job: fixed a bug that caused a state changes even in case of errors during apply ([#412](https://github.com/hashicorp/terraform-provider-nomad/pull/412))
```
gulducat, line 14: super nit:

```suggestion
* resource/nomad_job: fixed a bug that caused a state change even in case of errors during apply ([#412](https://github.com/hashicorp/terraform-provider-nomad/pull/412))
```

or

```suggestion
* resource/nomad_job: fixed a bug that caused state changes even in case of errors during apply ([#412](https://github.com/hashicorp/terraform-provider-nomad/pull/412))
```

nomad/resource_job.go
```diff
@@ -409,6 +410,10 @@ func resourceJobRegister(d *schema.ResourceData, meta interface{}) error {
 		return fmt.Errorf("error applying jobspec: %s", err)
 	}
 
+	if !d.IsNewResource() {
+		d.Partial(false)
```
gulducat, line 414: This is after the job is successfully registered, but before its deployment has succeeded (assuming `detach = false`).  I suppose this makes sense, because the resource is a job, not a deployment, but I guess I want to flag that a failing deployment would still show "no changes" on a subsequent apply, right? So in that case, one would need to `taint` or otherwise destroy and re-apply the job, if the fix for the failing deployment was something other than the job resource config.

gulducat (approved):
:partly_sunny: (this is a "partly sunny" emoji :yum:)

CHANGELOG.md
lgfa29: Ops, I meant the second one. Thanks!

nomad/resource_job.go
lgfa29: That's right, but that's Nomad's behaviour. The goal of the TF provider is to update the job state in Nomad. Once that's complete Nomad is responsible for the deployment. Applying again has not effect the same way as running `nomad job run` would also not have any effect.

PR #419, opened 2024-01-10
Update csi_volume_registration.html.markdown

attachmentgenie:
using correct resource name, instead of deprecated resource


lgfa29 (approved):
Ops, good catch! Thank you for the fix @attachmentgenie!

Issue #417, opened 2024-01-07
nomad_csi_volume missing context block


CarbonCollins:
Hi there,

Thank you for opening an issue. Please note that we try to keep the Terraform issue tracker reserved for bug reports and feature requests. For general usage questions, please see: https://www.terraform.io/community.html.

### Terraform Version

Terraform v1.6.6
on linux_amd64
+ provider registry.terraform.io/hashicorp/nomad v2.1.0
+ provider registry.terraform.io/hashicorp/vault v3.14.0

### Nomad Version
1.7.2

### Provider Configuration
Which values are you setting in the provider configuration? 

```hcl
provider "nomad" {}
```

### Environment Variables
Do you have any Nomad specific environment variable set in the machine running Terraform?

```sh
NOMAD_REGION=se
NOMAD_NAMESPACE=c3-networking
NOMAD_ADDR={REDACTED}
NOMAD_TOKEN={REDACTED}
```

### Affected Resource(s)
Please list the resources as a list, for example:
- nomad_csi_volume
- nomad_csi_volume_registration

### Terraform Configuration Files
```hcl
# Copy-paste your Terraform configurations here - for large Terraform configs,
# please use a service like Dropbox and share a link to the ZIP file. For
# security, you can also encrypt the files using our GPG public key.
```

### Expected Behavior
be able to define context block on the nomad_csi_volume resource like you can with the  nomad_csi_volume_registration

### Actual Behavior
You are unable to define the context block on the nomad_csi_volume resource

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. `terraform apply`

### Important Factoids
There is nothing atypical but when I tried to deploy my job to one of the volumes created and registered via the terraform provider my Nomad job wont start with the error:
```
failed to setup alloc: pre-run hook "csi_hook" failed: mounting volumes: rpc error: code = InvalidArgument desc = unknown/unsupported node_attach_driver: undefined
``` 

### References
context docs in (csi_volume_registration - context)[https://registry.terraform.io/providers/hashicorp/nomad/latest/docs/resources/csi_volume_registration#context]

CarbonCollins:
Just as some extra info on this one, I tried crating the volume via a volume specification file with the same information that the nomad_csi_volume resource has and It successfully created without returning the unsupported node_attach_driver error.

lgfa29:
Hi @CarbonCollins 👋 

[...]

PR #421, opened 2024-01-19
Add context to resource csi_volume

clarkbains:
Solves #417


lgfa29:
Thanks for the PR @clarkbains!

I made a comment on #417 for more information because, according to the [docs](https://developer.hashicorp.com/nomad/docs/other-specifications/volume#differences-between-create-and-register), `context` should only be set on volume registration, not creation.

Did you hit any problems because of this as well?

PR #429, opened 2024-02-07
Adding support for OIDCDisableUserInfo

netdata-be:
In nomad 1.7.3 a new config has been [introduced](https://github.com/hashicorp/nomad/pull/19566) Adding support for this config parameter


lgfa29 (approved):
Thank you for the contribution @netdata-be!

I pushed a commit to fix a conflict in `go.mod` and also to pin the `nomad/api` dependency to the `v1.7.3` tag.

PR #434, opened 2024-02-27
Add support for filtering namespace in allocations resource, as this defaults to "default"

MatthewJohn:


lgfa29 (approved):
Thanks for the PR @MatthewJohn!

I pushed a commit with a minor fix and some docs and a changelog entry for the change.

PR #441, opened 2024-03-12
test: fix `TestResourceVariable_namespaceChange`

### LINKED ISSUE #440: test: `TestResourceVariable_namespaceChange` failing
Test failure:
```
=== RUN   TestResourceVariable_namespaceChange
    testing_new.go:85: Error running post-test destroy, there may be dangling resources: exit status 1
        
        Error: error deleting namespace "var-test-namespace": Unexpected response code: 500 (namespace "var-test-namespace" contains at least one variable. All variables in namespace must be deleted before it can be deleted)
```

lgfa29:
Establish implicit dependency between the variable and the new namespace so they are destroyed in the right order.

Closes #440

### DIFF
--- a/nomad/resource_variable_test.go
+++ b/nomad/resource_variable_test.go
@@ -98,7 +98,7 @@ resource nomad_namespace "nomad_var_test" {
   name = "%s"
 }
 %s
-`, namespace, testResourceVariable_initialConfig(namespace, path))
+`, namespace, testResourceVariable_initialConfig("${nomad_namespace.nomad_var_test.name}", path))
 }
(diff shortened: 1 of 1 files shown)


gulducat (approved):
I do TF provider stuff so rarely, weird to see hcl variable syntax as a golang string argument, but after a bit of mental acrobatics through `Sprintf()`s, LGTM!

lgfa29:
Yeah, Terraform tests run Terraform configs, so using resource reference is how you would normally write this to make sure the DAG is properly defined.

PR #442, opened 2024-03-12
acl_auth_method: fix token name format default

lgfa29:
Set the token name default value used by Nomad to avoid unnecessary plan changes when the attribute is not defined.

Reported in https://github.com/hashicorp/terraform-provider-nomad/pull/403#issuecomment-1960136907


shoenig (approved):

lgfa29:
Failing test fixed in https://github.com/hashicorp/terraform-provider-nomad/pull/441

PR #447, opened 2024-03-28
Add JWKS data source for fetching public keys for JWT validation

clf-cklf:
This adds a `data` provider for the JWKS public keys used for verifying workload identity JWTs. These are exposed at the `/.well-known/jwks.json` HTTP API endpoint.

One use for this would be to provide these keys to Vault via https://registry.terraform.io/providers/hashicorp/vault/latest/docs/resources/jwt_auth_backend#jwt_validation_pubkeys. This would be useful, because Vault cannot access the HTTPS API with mTLS enabled unless one turns off client verification: https://developer.hashicorp.com/nomad/docs/integrations/vault/acl#mutual-tls-in-nomad

The keys are returned as both a list of objects with human-readable field names, and as a list of PEM-encoded X.509 public keys; the latter is, I believe, the preferred way to pass PKI information in Terraform, and is the format that the Terraform Vault provider expects.

### DIFF
--- a/nomad/data_source_jwks.go
+++ b/nomad/data_source_jwks.go
@@ -118,7 +118,8 @@ func dataSourceJWKSRead(d *schema.ResourceData, meta any) error {
 
 func keyToPem(key Key) (string, error) {
 
-	// FIXME does Nomad always use RSA keys for JWKS?
+    // Nomad also supports EdDSA keys, but they are not used for OIDC so only
+    // RSA keys should be listed in this endpoint.
 	if key.KeyType != "RSA" {
 		return "", fmt.Errorf("Key type not supported: %s", key.Algorithm)
 	}
--- a/nomad/data_source_jwks.go
+++ b/nomad/data_source_jwks.go
@@ -6,7 +6,6 @@ package nomad
 import (
 	"fmt"
 	"log"
-	"strings"
 
 	"crypto/rsa"
 	"crypto/x509"
@@ -155,10 +154,7 @@ func keyToPem(key Key) (string, error) {
 			Bytes: x509Cert,
 		})
 
-	// Vault renders the keys without a trailing newline; strip this out to avoid unneccesary updates
-	// FIXME this might be more appropriate to handle as a `chomp` on the terraform side when
-	// interacting with Vault. If so, document this.
-	return strings.TrimSpace(string(x509CertEncoded)), nil
+	return string(x509CertEncoded), nil
 }
 
 func fromKeys(keys []Key) []interface{} {
--- a/nomad/data_source_jwks.go
+++ b/nomad/data_source_jwks.go
@@ -1,0 +1,178 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+package nomad
+
+import (
+	"fmt"
+	"log"
+	"strings"
+
+	"crypto/rsa"
+	"crypto/x509"
+	"encoding/base64"
+	"encoding/pem"
+	"math/big"
+
+	"github.com/hashicorp/nomad/api"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/id"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+func dataSourceJWKS() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceJWKSRead,
+		Schema: map[string]*schema.Schema{
+			"keys": {
+				Description: "JSON Web Key Set (JWKS) public keys for validating workload identity JWTs",
+				Type:        schema.TypeList,
+				Computed:    true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"key_use": {
+							Type:     schema.TypeString,
+							Computed: true,
+						},
+						"key_type": {
+							Type:     schema.TypeString,
+							Computed: true,
+						},
+						"key_id": {
+							Type:     schema.TypeString,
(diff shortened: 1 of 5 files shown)


nomad/data_source_jwks.go
```diff
@@ -0,0 +1,178 @@
...
+	if err := d.Set("pem_keys", pemKeys); err != nil {
+		return fmt.Errorf("error setting JWKS pemKeys: %s", err)
+	}
+
+	return nil
+}
+
+func keyToPem(key Key) (string, error) {
+
+	// FIXME does Nomad always use RSA keys for JWKS?
```
clintonford-btnl, line 122: I'm not sure about this, but I think Nomad only uses RSA for this?

nomad/data_source_jwks.go
```diff
@@ -0,0 +1,178 @@
...
+	}
+
+	x509CertEncoded := pem.EncodeToMemory(
+		&pem.Block{
+			Type:  "PUBLIC KEY",
+			Bytes: x509Cert,
+		})
+
+	// Vault renders the keys without a trailing newline; strip this out to avoid unneccesary updates
+	// FIXME this might be more appropriate to handle as a `chomp` on the terraform side when
```
clintonford-btnl, line 159: I'm not sure if it is appropriate to do data clean-up in this code to make it compatible with another provider, especially as it can be handled easily in Terraform itself. If it is ok to do this, I can remove this comment. If not, I can remove the `TrimSpace`.

wraithm:
What do you think about this @lgfa29 ?

nomad/data_source_jwks.go
lgfa29: In theory the root key can also use `EdDSA`, but I think for OIDC RSA is required.
https://github.com/hashicorp/nomad/blob/78f9f178670097eac9036c4bf83f7e0ceb55e6f3/nomad/encrypter.go#L499-L505

But we can probably remove this check altogether? It would also help future-proof this data source in case new algorithms are added later on.

clintonford-btnl:
Thanks @lgfa29 ! I'm on vacation this week but will be able to respond and implement your suggestions this coming Monday. Your comments all look good, no concerns from me at this point.

nomad/provider.go
```diff
@@ -163,6 +163,7 @@ func Provider() *schema.Provider {
 			"nomad_regions":          dataSourceRegions(),
 			"nomad_volumes":          dataSourceVolumes(),
 			"nomad_variable":         dataSourceVariable(),
+			"nomad_jwks":             dataSourceJWKS(),
```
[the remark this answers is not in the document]
clintonford-btnl: Fixed - thank you!

nomad/data_source_jwks.go
clintonford-btnl: I did a bit of investigation on this: to support EdDSA it looks like we would need to add `crv` and `x` fields for `Key`s, and then construct a `ed25519` public key from them. We would still need the `KeyType` check to see which type of key to construct ("OKP" for EdDSA).

Also, as you pointed out, OIDC doesn't support non-RSA keys, so Nomad doesn't have code to make JWKs out of them. We would therefore need to construct a mocked-up key to test with.

I'm happy to do this, just want to make sure this sounds like worthwhile effort at this point. An alternative approach for future-proofing might be to ask for Nomad's `ListKeys` RPC to be make available via the HTTP API and fetch the keyset directly.

What do you think?

lgfa29: Ah, I see. I think we can just leave RSA support like it is then, maybe just update the comment to explain why only RSA is supported?

Supporting new key types seem like will require code changes here as well, so not a bit deal.

nomad/data_source_jwks.go
lgfa29: Yeah, using Terraform functions to fix this seems reasonable 👍

nomad/data_source_jwks.go
```diff
@@ -0,0 +1,174 @@
...
+	if err := d.Set("pem_keys", pemKeys); err != nil {
+		return fmt.Errorf("error setting JWKS pemKeys: %s", err)
+	}
+
+	return nil
+}
+
+func keyToPem(key Key) (string, error) {
+
+	// FIXME does Nomad always use RSA keys for JWKS?
```
lgfa29, line 121:
```suggestion
    // Nomad also supports EdDSA keys, but they are not used for OIDC so only
    // RSA keys should be listed in this endpoint.
```

I wanted to push a quick fix to this comment, but it doesn't seem like I have write access to the PR 😅 

Here's what I was thinking of writing. There's also a conflict in `CHANGELOG.md` that needs to be fixed before we can merge it.

nomad/data_source_jwks.go
clintonford-btnl: Looks like you and @wraithm already added a comment - thank you! I pushed a quick fix to change the indentation to match the surrounding code, I think it looks good to go now?

lgfa29 (approved):
Thank you for the updates @clf-cklf.

I no longer work at HashiCorp, so I can't get this PR merged, but I'm leaving my approval here to signal to the rest of the team that these changes look good 😄

wraithm:
Thanks @lgfa29! Do you know who's taken over maintainership of this project?

jrasell (approved):
LGTM, thanks @clf-cklf.

clf-cklf:
> Thank you for the updates @clf-cklf.
> 
> I no longer work at HashiCorp, so I can't get this PR merged, but I'm leaving my approval here to signal to the rest of the team that these changes look good 😄

Thanks @lgfa29 ! Best wishes for your next role!

PR #448, opened 2024-04-02
Add JWT options for acl_auth_method

### LINKED ISSUE #422: [Feature Request] Add JWT type to nomad_acl_auth_method
The `nomad_acl_auth_method` resource currently only has support for OIDC type methods. It would be great, if the JWT type would also be supported (along with the added parameters in the config block).

jorgemarey:
Fixes #422

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## 2.2.1 (Unreleased)
 
+IMPROVEMENTS:
+* resource/acl_auth_method: add support for configuring a JWT auth-method ([#448](https://github.com/hashicorp/terraform-provider-nomad/pull/448))
+
 ## 2.2.0 (March 12, 2024)
 
 IMPROVEMENTS:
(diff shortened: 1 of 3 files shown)


lgfa29:
Thanks for the PR @jorgemarey!

Have you tested what happens if `type` changes? Would we leave an orphaned auth method in Nomad? Maybe we would need to update it to have `ForceNew: true`?

I was also wondering if we should validate `config` based on `type`, but I don't think we can do cross-field validation at plan time, at which point we may as well let Nomad knows if something is wrong on apply.

Would you also be able to add a CHANGELOG entry and update the [website docs](https://github.com/hashicorp/terraform-provider-nomad/blob/main/website/docs/r/acl_auth_method.html.markdown)?

Thanks!

jorgemarey:
Hi @lgfa29 

> Have you tested what happens if `type` changes? Would we leave an orphaned auth method in Nomad? Maybe we would need to update it to have `ForceNew: true`?

I didn't notice this, but I just tested and the method seems to change correctly in nomad. So I guess there's no need to set ForceNew to true then.


> I was also wondering if we should validate `config` based on `type`, but I don't think we can do cross-field validation at plan time, at which point we may as well let Nomad knows if something is wrong on apply.

I don't know how to do this. I you think we can do something on plan time let me know and I'll make the change.

> Would you also be able to add a CHANGELOG entry and update the [website docs](https://github.com/hashicorp/terraform-provider-nomad/blob/main/website/docs/r/acl_auth_method.html.markdown)?

Done!

lgfa29 (approved):
Looks great, thank you @jorgemarey!

PR #462, opened 2024-05-29
deps: update Nomad to v1.8.0

### LINKED ISSUE #458: Nomad 1.8 support
With Nomad 1.8, we need to update the terraform provide to bump our dependences and also make changes to handle some new blocks such as transparent_proxy, ui config (description, urls), an "insecure" field in artifact{}

Related: https://github.com/hashicorp/terraform-provider-nomad/pull/454

jrasell:
closes #458


go.mod
```diff
@@ -2,13 +2,15 @@ module github.com/hashicorp/terraform-provider-nomad
...
+replace (
+	// Set to the version used by Nomad
+	// https://github.com/hashicorp/nomad/blob/v1.6.0-rc.1/go.mod#L74
+	github.com/hashicorp/hcl/v2 => github.com/hashicorp/hcl/v2 v2.9.2-0.20220525143345-ab3cae0737bc
 
-// Pin Nomad to a commit where jobspec and jobspec2 are lincesed under MPL-2.0.
-// Remove when updating to Nomad 1.7.3+
-replace github.com/hashicorp/nomad => github.com/hashicorp/nomad v0.0.0-20240129172537-56ea3f7cdafd
+    // Fix error tidying due to Nomad downstream dependencies and the recent
+    // migration of the metrics library.
```
gulducat, line 11: bit of inconsistent spacing here

```suggestion
	// Fix error tidying due to Nomad downstream dependencies and the recent
	// migration of the metrics library.
```

I think that gets em to all be tabs?

nomad/resource_job_test.go
```diff
@@ -3272,6 +3328,37 @@ job "foo-multiregion" {
 }
 `
 
+var testResourceJobScheduleBlock = `
```
gulducat, line 3331: hey neat!  if you're adding extra testing for this, do we also want to cover the new [`ui{}` block](https://developer.hashicorp.com/nomad/docs/job-specification/ui)?

`raw_exec` also has the new [cgroup overrides](https://developer.hashicorp.com/nomad/docs/drivers/raw_exec#cgroup_v1_override)...

how much do we test that kind of thing here in the TF provider?

gulducat (approved):
thank you!

nomad/resource_job_test.go
gulducat: I guess the raw_exec cgroup stuff is part of the driver config, which probably(?) doesn't get parsed here(?)

jrasell: I usually add a test which at least covers the version, but the UI makes sense to add, so this can be found in: https://github.com/hashicorp/terraform-provider-nomad/pull/462/commits/67d75b858ab972c08d1b6d81646a258eb30a2986

go.mod
jrasell: Fixed in: https://github.com/hashicorp/terraform-provider-nomad/pull/462/commits/67d75b858ab972c08d1b6d81646a258eb30a2986

PR #470, opened 2024-07-30
namespaces: handle quota spec disassociation and non-terminal allocations

pkazmierczak:
Nomad 1.8.2 will [no longer allow](https://github.com/hashicorp/nomad/pull/23499) for deleting of quota specs if they are associated with a namespace. Namespace deletion will now also detect non-terminal allocations, so we have to support retries when deleting these. 

Fixes https://github.com/hashicorp/nomad/issues/23672
Internal ref: https://hashicorp.atlassian.net/browse/NET-10579

_note to reviewers:_ this PR also bumps the version of Nomad used for unit tests to 1.8.2. I will make a follow-up PR that installs nomad with hc-install so that we can assure we're always running tests against latest stable.

### DIFF
--- a/nomad/resource_namespace_test.go
+++ b/nomad/resource_namespace_test.go
@@ -227,10 +227,24 @@ resource "nomad_namespace" "test" {
 
 func testResourceNamespace_configWithQuota(name, quota string) string {
 	return fmt.Sprintf(`
+resource "nomad_quota_specification" "test_quota" {
+  name        = "%[2]s"
+  description = "A Terraform acctest quota spec"
+
+  limits {
+    region = "global"
+
+    region_limit {
+      cpu       = 2400
+      memory_mb = 1200
+    }
+  }
+}
+
 resource "nomad_namespace" "test" {
-  name = "%s"
+  name = "%[1]s"
   description = "A Terraform acctest namespace"
-  quota = "%s"
+  quota = "%[2]s"
 
   meta = {
     key = "value",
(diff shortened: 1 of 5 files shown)


nomad/resource_quota_specification.go
```diff
@@ -110,9 +110,28 @@ func resourceQuotaSpecificationDelete(d *schema.ResourceData, meta interface{})
...
+	}
+	for _, ns := range nss {
+		if ns.Quota == name {
+			// dissasociate the quota with its ns
+			ns.Quota = ""
+			_, err := client.Namespaces().Register(ns, nil)
+			if err != nil {
+				return fmt.Errorf(
+					"error disassociating quota spec %q with namespace %s: %s",
+					name, ns.Name, err.Error(),
```
tgross, line 126:
```suggestion
					"error disassociating quota spec %q with namespace %s: %w",
					name, ns.Name, err,
```

nomad/resource_namespace_test.go
```diff
@@ -266,6 +326,21 @@ func testResourceNamespace_initialCheck(name string) resource.TestCheckFunc {
 	}
 }
 
+func testResourceNamespaceWithQuota_check(name, quota string) resource.TestCheckFunc {
+	return func(s *terraform.State) error {
+		client := testProvider.Meta().(ProviderConfig).client
+		namespace, _, err := client.Namespaces().Info(name, nil)
+		if err != nil {
+			return fmt.Errorf("error reading back namespace %q: %s", name, err)
```
tgross, line 334:
```suggestion
			return fmt.Errorf("error reading back namespace %q: %w", name, err)
```

tgross (approved):
LGTM!

PR #483, opened 2024-10-15
provider: update Nomad dependency to 1.9.0

jrasell:
The update removes support for `hcl1` and the `hcl2.enabled` flag. This is because Nomad 1.9 removes hcl1 support.


tgross (approved):
LGTM!

PR #497, opened 2025-01-10
testing: reduce skipped checks for non-enterprise builds

tgross:
Namespaces were moved from Enterprise to CE some time ago, but it appears that some of the tests for the provider never had their ENT-only gate removed so they're not being run on CE. One test for scheduler configuration was missing an ENT-only gate as well.

Additionally, missing `t.Helper()` flags were causing these skips to be reported as if they were version number check failures, which was confusing. This changeset fixes that and simplifies the test helper for version checks so the helper function that gets the version from the node returns a value rather than taking a callback, and uses the `Core` method on the checked version to allow easier development against new features on Nomad's main branch.


pkazmierczak (approved):
👍

PR #498, opened 2025-01-13
dynamic host volumes

tgross:
Adds support for dynamic host volumes, which will be shipping in Nomad 1.10.0. This includes a resource for creating volumes, another for registering them, and a shared data source.

Ref: https://hashicorp.atlassian.net/browse/NET-11833


tgross:
Needs updated API package. Blocked on https://github.com/hashicorp/terraform-provider-nomad/pull/518

tgross:
Originally the plan here was to avoid merging so that we wouldn't have 1.10 API code in main before RC, but because we needed to do that to test backcompat changes for vault/consul tokens, I'm going to propose we merge this in now. We'll make the final library update to the GA `api` package when we ship.

CHANGELOG.md
```diff
@@ -5,6 +5,8 @@ BREAKING CHANGES:
 * resource/nomad_job: The deprecated `consul_token` and `vault_token` fields have been removed ([#518](https://github.com/hashicorp/terraform-provider-nomad/pull/518)
 
 IMPROVEMENTS:
+* **New Resources**: `nomad_dynamic_host_volume_create` and `nomad_dynamic_host_volume_register` to create and register dynamic host volumes ([#498](https://github.com/hashicorp/terraform-provider-nomad/pull/498))
```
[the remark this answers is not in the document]
gulducat: Still needa update this markdown too, eh?

Also in the tests

nomad/resource_dynamic_host_volume_test.go
```diff
@@ -0,0 +1,291 @@
...
+			},
+		},
+		CheckDestroy: testResourceDynamicHostVolume_checkDestroy(
+			testResourceNameDynamicHostVolume),
+	})
+}
+
+func testResourceDynamicHostVolume_config(name string) string {
+	return fmt.Sprintf(`
+resource "nomad_dynamic_host_volume_create" "test" {
```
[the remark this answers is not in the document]
tgross: Should be 1.9.7-dev, but yeah. Somehow I missed these were getting skipped. Lemme rework that.

tgross:
Should be good-to-go now.

nomad/resource_dynamic_host_volume_registration.go
```diff
@@ -0,0 +1,205 @@
...
+
+	parameters := make(map[string]string)
+	for name, value := range d.Get("parameters").(map[string]any) {
+		parameters[name] = value.(string)
+	}
+
+	capacity, err := humanize.ParseBytes(d.Get("capacity").(string))
+	if err != nil {
+		return fmt.Errorf("could not parse capacity value as bytes: %w", err)
+	}
```
gulducat, line 169: I hit this error condition if "capacity" is empty.  minimal repro:

```hcl
resource "nomad_dynamic_host_volume" "one" {
  name      = "volume-one"
  plugin_id = "mkdir"
  namespace = "default"
  capability {
  }
}

resource "nomad_dynamic_host_volume_registration" "two" {
  # capacity = "5mb" # optional
  name      = "volume-two"
  node_id   = nomad_dynamic_host_volume.one.node_id
  host_path = nomad_dynamic_host_volume.one.host_path
  capability {
  }
}
```

> Error: could not parse capacity value as bytes: strconv.ParseFloat: parsing "": invalid syntax

nomad/resource_dynamic_host_volume_registration_test.go
```diff
@@ -0,0 +1,211 @@
...
+  }
+
+  capability {
+    access_mode     = "single-node-reader-only"
+    attachment_mode = "file-system"
+  }
+
+  parameters = {
+    some_key = "some_value"
+  }
```
gulducat, line 109: tangential to TF, but isn't it weird to have `parameters` in a registration? that's entirely for plugins, isn't it? and registration doesn't use plugins.

nomad/data_source_dynamic_host_volume.go
```diff
@@ -0,0 +1,205 @@
...
+	}
+
+	d.SetId(id)
+	return nil
+}
+
+// getDynamicHostVolumeNamespacedID returns the namespace and ID for a dynamic
+// host volume from the resource data
+func getDynamicHostVolumeNamespacedID(d *schema.ResourceData) (string, string) {
+	id := d.Id()
```
gulducat, line 198: this causes an error in the `data` source, because it's empty until there's state?

```hcl
resource "nomad_dynamic_host_volume" "one" {
  name      = "volume-one"
  plugin_id = "mkdir"
  namespace = "default"
  capability {
  }
}

data "nomad_dynamic_host_volume" "one" {
  id = nomad_dynamic_host_volume.one.id
}
```

> Error: Failed to get information about "": Unexpected response code: 404 (resource not found)

if I do this, then it fixes it, but I haven't checked to see if there are any other implications of doing so

```suggestion
func getDynamicHostVolumeNamespacedID(d *schema.ResourceData) (string, string) {
	id := d.Get("id").(string)
```

gulducat (changes requested):
I haven't fine-tooth-combed the docs yet, but have a couple nits, and found a couple bugs.

nomad/resource_dynamic_host_volume.go
```diff
@@ -0,0 +1,388 @@
...
+			"capacity_max": {
+				Description: "Requested maximum capacity",
+				Type:        schema.TypeString,
+				Optional:    true,
+			},
+			"capacity_min": {
+				Description: "Requested minimum capacity",
+				Type:        schema.TypeString,
+				Optional:    true,
+			},
```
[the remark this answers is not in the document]
gulducat: to be clear, these could-validate-sooner nits are not blocking, as the API prevents anything actually bad happening. they're just nice-to-haves.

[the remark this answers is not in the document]
tgross: Unfortunately it doesn't look like the [`ValidateDiagFunc` hooks](https://developer.hashicorp.com/terraform/plugin/sdkv2/schemas/schema-behaviors#validatediagfunc) allow you to do cross-field validation like that. We could validate during resource creation and return an error, but the user experience is roughly the same (and we'd have to keep validation in sync).

nomad/resource_dynamic_host_volume.go
```diff
@@ -0,0 +1,388 @@
...
+							Description: "The operator to use for comparison",
+							Type:        schema.TypeString,
+							Optional:    true,
+							Default:     "=",
+						},
+					},
+				},
+			},
+			"node_id": {
+				Description: "Node ID",
```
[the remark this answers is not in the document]
tgross: Yeah I like the idea of adding `ForceNew`. Will do.

nomad/data_source_dynamic_host_volume.go
tgross: This got refactored a bunch and I don't think I re-tested the data source. Lemme give that a try and report back.

nomad/resource_dynamic_host_volume_registration_test.go
tgross: The API supports it but you're right, it just becomes silent metadata in the register workflow. We could remove it from the API by stripping it out in the Register RPC, but the object is always going to have the fields available. Maybe we should drop it from the TF provider? We should probably update the [parameters](https://developer.hashicorp.com/nomad/docs/v1.10.x/other-specifications/volume/host#parameters) docs to say those only apply to the create workflow as well (although it's hinted at by talking about plugins).

nomad/resource_dynamic_host_volume.go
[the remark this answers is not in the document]
tgross: Done

nomad/resource_dynamic_host_volume_registration_test.go
tgross: On a second pass, I think we'll want to keep them in the provider, because the data source could be pulling from a volume that was created with either a register or create workflow and we can't know which. We could probably have a separate read-back method for the register resource, but then we'd end up with a ton of duplicated code to drift or a weird "don't include this one" parameter on the read path. What do you think?

Also, ironically, when running these against the rc.1 build, I had to remove the `parameter` block from our test configs for the create workflow, because that's the `mkdir` plugin which will now reject those.

gulducat: you can keep em in there if you want, so long as they're valid. I switched `foo = "bar"`-type stuff to `mode = "0755"` in some tests in `nomad` repo.

nomad/data_source_dynamic_host_volume.go
tgross: Ok, checked the current build end-to-end and that seems to work.

nomad/resource_dynamic_host_volume_registration_test.go
tgross: :facepalm: of course! Fixed.

gulducat (approved):
lgtm! :tada:

PR #502, opened 2025-01-16
testing: fix test flake in data.nomad_allocations test

tgross:
The acceptance test for `data.nomad_allocations` creates a job and then reads the allocations in the same configuration. This results in frequent test flakes because registering a job via `resource.nomad_job` will typically return before the allocations have actually been created, making the `depends_on` useless.

This frankly isn't a good workflow to encourage, as allocations are ephemeral. But fix the test by creating the job first, waiting for the allocations to be live, and then adding the `data.nomad_allocations` object to the config.

### DIFF
--- a/nomad/data_source_allocations_test.go
+++ b/nomad/data_source_allocations_test.go
@@ -4,14 +4,18 @@
 package nomad
 
 import (
+	"errors"
 	"fmt"
 	"regexp"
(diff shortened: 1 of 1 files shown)


jrasell (approved):
LGTM, thanks @tgross!

PR #516, opened 2025-03-17
bump: go version and x-repo deps

dduzgun-security:
Bumping the go version and x-repository dependencies.


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/hashicorp/terraform-provider-nomad
 
-go 1.23.6
+go 1.23.7
```
tgross, line 3: Why not bump to 1.24?

dduzgun-security: Good point, that's even better. https://github.com/hashicorp/terraform-provider-nomad/pull/516/commits/d394a5f5626ddefd70cfafe08e8fc5d53df2ed8f Wasn't too sure why we kept 1.23 but with a 1.24 toolchain and tried to make this PR to reduce possible impact/context behind

tgross (approved):
LGTM assuming CI is green

PR #518, opened 2025-03-26
provider: remove consul_token and vault_token

tgross:
In Nomad 1.10 the deprecated legacy Consul and Vault token workflow is being removed. Remove the `consul_token` and `vault_token` fields from the provider.

This also removes all Consul/Vault specific tests from this provider repo. With the tokens being removed, any tests of these features would be a test of Nomad and not the provider.


tgross:
You're right that should've been 3.0. Just :facepalm: on my part when we released, as Nomad core increments the `1.x.0` version on the breaking changes and I'd mapped that across to the provider unthinking. My bad. We'll do better.

PR #519, opened 2025-03-26
Adds verbose logging option to auth methods

nickwales:
This updates the auth method with the verbose_logging option that was added in 1.9.6. 

The API version needs to be updated in go.mod but I cannot find the version between current and what is in main so this should either be merged with that or with the next release.

### DIFF
--- a/nomad/provider_test.go
+++ b/nomad/provider_test.go
@@ -48,11 +48,17 @@ func TestProvider_impl(t *testing.T) {
 
 var testProvider *schema.Provider
 var testProviders map[string]*schema.Provider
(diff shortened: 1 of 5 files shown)


tgross (commented):
We'll need to `go get -u github.com/hashicorp/nomad/api`, but currently that'll be broken because of the work pending in https://github.com/hashicorp/terraform-provider-nomad/pull/518. Once that's been merged we can rebase this.

tgross:
@nickwales can you rebase this on main?

tgross:
Ok, go ahead and try to rebase this again.

PR #520, opened 2025-04-01
Nomad upgrade to 1.10.0-beta.1

tgross:
This PR upgrades the Nomad test target to 1.10.0-beta.1, and fixes the problems resulting from that:

* In https://github.com/hashicorp/terraform-provider-nomad/pull/518 when we updated the API package, we missed a `go mod tidy` and missed some leftover Consul/Vault infrastructure bits.

* In Nomad 1.9.7 we started redacting the OIDC client secret field from the Auth Method API. Once we update the Nomad test target, this results in incorrect plan diffs. But we don't want to simply suppress the diff of this field, because we need to record the original unredacted value in state. Update the resource to record the original secret to state and then use the state to override the value we get from the API when performing the diff. Note that it's impossible with or without this patch to import an Auth Method without manual intervention, because we can't create the OIDC client secret from nothing.

Ref: https://github.com/hashicorp/nomad/pull/25328


nomad/resource_acl_auth_method.go
```diff
@@ -264,21 +279,54 @@ func resourceACLAuthMethodRead(d *schema.ResourceData, meta interface{}) error {
...
 		return fmt.Errorf("error reading ACL Auth Method %q: %s", authMethodName, err.Error())
 	}
 	log.Printf("[DEBUG] Read ACL Auth Method %q", authMethod.Name)
 
+	unredactResource(d, authMethod, nil)
+	setStateFromResource(d, authMethod)
+	return nil
+}
+
+func fetchResource(client *api.Client, name string) (*api.ACLAuthMethod, error) {
```
gulducat, line 293: the package here is `nomad` (the whole provider) so I think this method is named too broadly for its namespace. ditto the other funcs below.

nomad/resource_acl_auth_method.go
```diff
@@ -264,21 +279,54 @@ func resourceACLAuthMethodRead(d *schema.ResourceData, meta interface{}) error {
...
+	log.Printf("[DEBUG] Reading ACL Auth Method %q", name)
+	authMethod, _, err := client.ACLAuthMethods().Get(name, nil)
+	if err != nil {
+		return nil, fmt.Errorf("error reading ACL Auth Method %q: %w", name, err)
+	}
+	log.Printf("[DEBUG] Read ACL Auth Method %q", name)
+	return authMethod, nil
+}
+
+func unredactResource(d *schema.ResourceData, fetchedAuthMethod, originalAuthMethod *api.ACLAuthMethod) {
```
gulducat, line 303: re:
> Note that it's impossible with or without this patch to import an Auth Method without manual intervention, because we can't create the OIDC client secret from nothing.

In my exp with Terraform, I wouldn't say this kind of thing is the norm, but it's not terribly uncommon to anticipate some extra hand-holding during a TF import.

I suppose the symptom would be that the very first `apply` after the import would show a diff, even if the secret in the user's HCL is actually identical to what Nomad has squirreled away in its state, yes?

gulducat (approved):
I would like more-specific func names, but LGTM!

nomad/resource_acl_auth_method.go
tgross: Oof, totally right. I was getting file tunnel vision

nomad/resource_acl_auth_method.go
tgross: Yeah, because the TF state would be `"redacted"`

gulducat (approved):
thanks!

PR #523, opened 2025-04-15
nomad_acl_auth_method: add pkce and client assertions

gulducat:
Add support for enabling PKCE and using client assertions (private key JWT) instead of client secrets.

* PR that introduced these features to Nomad: hashicorp/nomad#25231
* API documentation: https://developer.hashicorp.com/nomad/api-docs/v1.10.x/acl/auth-methods#oidcclientassertion

Also scope creeped a bit for a couple little things that were bothering me.

### DIFF
--- a/nomad/resource_acl_auth_method.go
+++ b/nomad/resource_acl_auth_method.go
@@ -82,11 +82,17 @@ func resourceACLAuthMethodConfig() *schema.Resource {
 	return &schema.Resource{
 		Schema: map[string]*schema.Schema{
 			"jwt_validation_pub_keys": {
-				Description:  "List of PEM-encoded public keys to use to authenticate signatures locally.",
-				Type:         schema.TypeList,
-				Elem:         &schema.Schema{Type: schema.TypeString},
-				Optional:     true,
-				ExactlyOneOf: []string{"config.0.jwks_url", "config.0.oidc_discovery_url"},
+				Description: "List of PEM-encoded public keys to use to authenticate signatures locally.",
+				Type:        schema.TypeList,
+				Elem:        &schema.Schema{Type: schema.TypeString},
+				Optional:    true,
+				ExactlyOneOf: []string{
+					// type = "JWT" options
+					"config.0.jwt_validation_pub_keys",
+					"config.0.jwks_url",
+					// type = "OIDC"
+					"config.0.oidc_discovery_url",
+				},
 			},
 			"jwks_url": {
 				Description: "JSON Web Key Sets url for authenticating signatures.",
@@ -99,10 +105,13 @@ func resourceACLAuthMethodConfig() *schema.Resource {
 				Optional:    true,
 			},
 			"oidc_discovery_url": {
-				Description:  "The OIDC Discovery URL, without any .well-known component (base path).",
-				Type:         schema.TypeString,
-				Optional:     true,
-				RequiredWith: []string{"config.0.oidc_client_id", "config.0.oidc_client_secret"},
+				Description: "The OIDC Discovery URL, without any .well-known component (base path).",
+				Type:        schema.TypeString,
+				Optional:    true,
+				RequiredWith: []string{
+					"config.0.oidc_discovery_url", // if this is set,
+					"config.0.oidc_client_id",     // client id must also be set
+				},
 			},
 			"oidc_client_id": {
 				Description: "The OAuth Client ID configured with the OIDC provider.",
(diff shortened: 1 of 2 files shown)


gulducat:
Pardon me, I still need to update the docs.

nomad/resource_acl_auth_method.go
```diff
@@ -300,21 +429,44 @@ func fetchACLAuthMethodResource(client *api.Client, name string) (*api.ACLAuthMe
...
+	// originalAuthMethod is what we just sent to Nomad, during resourceACLAuthMethod(Create|Update)
+	if originalAuthMethod != nil && originalAuthMethod.Config != nil {
+		if originalAuthMethod.Config.OIDCClientSecret != "" {
 			fetchedAuthMethod.Config.OIDCClientSecret = originalAuthMethod.Config.OIDCClientSecret
-			return
 		}
+		if originalAuthMethod.Config.OIDCClientAssertion != nil &&
+			originalAuthMethod.Config.OIDCClientAssertion.PrivateKey != nil &&
+			originalAuthMethod.Config.OIDCClientAssertion.PrivateKey.PemKey != "" {
+			fetchedAuthMethod.Config.OIDCClientAssertion.PrivateKey.PemKey = originalAuthMethod.Config.OIDCClientAssertion.PrivateKey.PemKey
```
tgross, line 450: is `fetchedAuthMethod.Config.OIDCClientAssertion` guaranteed to be non-nil if `fetchedAuthMethod.Config` is? It doesn't look like it based on the nil check we do further down.

tgross (commented):
This looks great! I don't have a OIDC setup handy to test it for myself this afternoon, but code is looking good.

nomad/resource_acl_auth_method.go
```diff
@@ -99,10 +111,13 @@ func resourceACLAuthMethodConfig() *schema.Resource {
...
-				Type:         schema.TypeString,
-				Optional:     true,
-				RequiredWith: []string{"config.0.oidc_client_id", "config.0.oidc_client_secret"},
+				Description: "The OIDC Discovery URL, without any .well-known component (base path).",
+				Type:        schema.TypeString,
+				Optional:    true,
+				RequiredWith: []string{
+					"config.0.oidc_discovery_url", // if this is set,
+					"config.0.oidc_client_id",     // client id must also be set
+				},
```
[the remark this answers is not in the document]
gulducat: A lot of the go docs on these fields have strained my brain cells a bit too, yeah. I took it to mean (and experimentation bears out) that "if I, myself, <my name here, explicitly> am set, then this other thing(s) must also be set."

So in total in this PR, from `RequiredWith`s:

* if `oidc_discovery_url` is set, `oidc_client_id` is required, too (defined here in `oidc_discovery_url`)
* if `oidc_client_secret` is set, `oidc_client_id` is required, too (defined below in `oidc_client_secret`)
* if `oidc_client_assertion` is set, `oidc_client_id` is required, too (defined below in `oidc_client_assertion`)

`oidc_discovery_url` is required by virtue of `ExactlyOneOf` up in `jwt_validation_pub_keys` (unless one of the jwt/jwks options are set).

I could not for the life of me find a way to say "one of: client secret, or assertion must be provided if client id (or discovery url?) is set, otherwise we must be in jwt-land, so don't worry about it."  And I think _technically_, one can have an unsecured OIDC setup without any secret material, so I opted to stop banging my head against that problem, and let reality sort it out.

nomad/resource_acl_auth_method.go
gulducat: anecdotally, it doesn't get tripped in any tests, but I still like to guard against `nil`, so I'll do some extra checkin :+1:

nomad/resource_acl_auth_method.go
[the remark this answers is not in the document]
tgross: Yeah, it's nice to give as much validation as we can but at some point you have to worry about drift anyways. LGTM

nomad/resource_acl_auth_method.go
gulducat: the pile of conditions was feeling like a jenga tower, so I made a (pretty precious) wrapper to make this stretch of code easier to read (and safe). I could be convinced to pull back from that, but it does the thing. whatcha think?

tgross: I'd have probably made those free functions but here in Terraform provider code it's not like we're instantiating this tramp object a million times in a server or something, so that's just a matter of taste, not function. LGTM

tgross (approved):
LGTM

PR #525, opened 2025-04-16
run ent tests on ent clusters

gulducat:
Ent feature tests have been getting skipped. At first I thought it was due to CI not downloading an ent binary, but no it turns out that this one line is stripping the `+ent` metadata from the version.

I don't know why `.Core()` was put here initially, but it smells like it might have been vestigial after some reworking in 0cab885b832dd9ddd43594d953fd3605cabe6904 -- one caller calls `.Core()` on the _result_ of this `testGetVersion()` func to compare against min version.

The tests do still pass (on my laptop) when run without an ent binary, too.

### DIFF
--- a/nomad/provider_test.go
+++ b/nomad/provider_test.go
@@ -110,7 +110,6 @@ func testGetVersion(t *testing.T) *version.Version {
 		if version, err := version.NewVersion(nodes[0].Version); err != nil {
 			t.Skip("could not parse node version: ", err)
 		} else {
-			version = version.Core()
 			return version
 		}
 	} else {


tgross (approved):
LGTM!

I suspect I left that call to `Core` while working with a pre-release binary for 1.10.0 before we shipped DHV. It might be worth checking that we're not making future development hit the same problem.

gulducat:
> a pre-release binary for 1.10.0

good call. I think we're in the clear:

```go
for _, vString := range []string{
	"1.10.0",
	"1.10.0+ent",
	"1.10.0-rc.1",
	"1.10.0-rc.1+ent",
	"1.10.1-dev",     // current CE main - `make dev`
	"1.10.1-dev+ent", // current ENT main
} {
	v, err := version.NewVersion(vString)
	if err != nil {
		panic(err)
	}
	fmt.Printf("core: %q; meta: %q\n", v.Core(), v.Metadata())
}
```

output:

```
core: "1.10.0"; meta: ""
core: "1.10.0"; meta: "ent"
core: "1.10.0"; meta: ""
core: "1.10.0"; meta: "ent"
core: "1.10.1"; meta: ""
core: "1.10.1"; meta: "ent"
```

PR #536, opened 2025-05-28
Add auth login jwt

### LINKED ISSUE #473: Feature request: Add support for login with externally issued tokens (JWT)
Since Nomad now has support for logging in with JWTs issued by external identity providers, and those [can be configured by this terraform provider](https://github.com/hashicorp/terraform-provider-nomad/pull/448), the next step would be to enable login with these tokens.

The endpoint to use is documented here: https://github.com/hashicorp/terraform-provider-nomad/pull/448

The configuration syntax could be either individual properties like, where they both have to be null or not-null:

```hcl
provider "nomad" {
  auth_method_name = "github"
  login_token      = var.login_token
}
```

or a like the [`auth_login_jwt`](https://registry.terraform.io/providers/hashicorp/vault/latest/docs#jwt) block in the vault provider, where both would be required if the block is defined:

```hcl
provider "nomad" {
  auth_login_jwt {
    auth_method_name = "github"
    login_token      = var.login_token
  }
}
```

jorgemarey:
Fixes #473


tgross (changes requested):
Hi @jorgemarey! I've built this locally and tested it, and overall this looks great. In addition to the small correct I've left, can you add documentation for these fields to `website/docs/index.html.markdown`?

Thanks for the PR!

jorgemarey:
Hi @tgross, thank you for reviewing this. I'll made the requested changes. 
Just wanted to add that we have been testing this in our environment for the last few weeks, and there's an issue that I think still needs to be addressed:
- When we are using a non authoritative region for the login, and the auth method creates a global token, sometimes we get a 403 when executing the request. I guess this is due to how nomad replication works: when a global token is created, that token is created first in the autoritative region and the the others replicate that value, so when the terraform resource call is made the token, sometimes, is not yet in the region.

The thing is that I don't know how to properly address this. We could made some request to check if the token is replicated, but we can't know for sure which permisions does the token have... We could just wait a few seconds, but that doesn't seem the best solution. Any ideas?

tgross:
> I guess this is due to how nomad replication works: when a global token is created, that token is created first in the autoritative region and the the others replicate that value

Oh, good catch! Replication between federated regions is asynchronous, so there's no guarantee the token exists in the non-authoritative region when the API call returns.

You could make a call to [Read Self Token](https://developer.hashicorp.com/nomad/api-docs/acl/tokens#read-self-token) to the target region. That'll return a 403 Permission Denied if the token doesn't exist. If it fails, you could do a short backoff-and-retry and give up after a little while to make sure TF isn't sitting there stuck if there's a cross-region replication failure.

googol:
Thanks @jorgemarey for implementing this!

PR #542, opened 2025-07-31
provider: add enabled_network_modes and disabled_network_modes

olljanat:
provider: add enabled_network_modes and disabled_network_modes

fixes #533

PR #543, opened 2025-07-31
ci: enable tests to pull requests

olljanat:
I noticed when investigated https://github.com/hashicorp/terraform-provider-nomad/pull/542#pullrequestreview-3075640813 that PR validation is not enabled at all in here.

You can see from my fork https://github.com/olljanat/terraform-provider-nomad/pull/1 that merging this change and doing PR close + reopen triggered tests.


gulducat (approved):
thank you!

Issue #548, opened 2025-08-21
Add support for `once` template parameter


bouncyboro:
### Terraform Version
```shell
terraform -v
Terraform v1.13.0
on linux_amd64
...
+ provider registry.terraform.io/hashicorp/nomad v2.5.0
...
```

### Nomad Version
```shell
nomad server members
Name                                    Address        Port  Status  Leader  Raft Version  Build   Datacenter  Region
...  ...  4648  alive   true    3             1.10.4  dev         global
...  ...  4648  alive   false   3             1.10.4  dev         global
...  ...  4648  alive   false   3             1.10.4  dev         global
``` 

### Provider Configuration
Which values are you setting in the provider configuration? 

```hcl
provider "nomad" {
  address   = ...
  region    = "global"
  secret_id = ...
}
```

### Environment Variables
Do you have any Nomad specific environment variable set in the machine running Terraform?

No.

### Affected Resource(s)
- nomad_job

### Terraform Configuration Files
```hcl
resource "nomad_job" ... {
  jobspec = templatefile("${path.module}/job.nomad", {
  ...
  })
}
```

job.nomad file:
```
job ... {
  ...
  group ... {
    ...
    task ... {
      ...
      template {
        env = true
        once = true

        data = ...

        destination = ...
      }
    }
  }
}
```

### Error Output
```
Error: error parsing jobspec: :275,9-13: Unsupported argument; An argument named "once" is not expected here.
```

### Expected Behavior
 `once` can be used in `template` block of a task.

### Actual Behavior
`terraform apply` fails with the above error.

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. Add `once` to template block of a task.
2. `terraform apply`

tristanmorgan:
Sorry to reopen this issue @tgross, but provider 2.5.1 (released September 24th) still reports the error.

```
╷
│ Error: error parsing jobspec: :320,9-13: Unsupported argument; An argument named "once" is not expected here.
│ 
│   with nomad_job.everything["alloy.nomad"],
│   on nomad.tf line 50, in resource "nomad_job" "everything":
│   50: resource "nomad_job" "everything" {
│ 
╵
```

PR #562, opened 2025-10-08
deps: update Nomad Go API package

tgross:
When we updated the Nomad package in 2.5.1, we forgot to update the Nomad API package as well, so using features like `template.once` are not working.

Fixes: https://github.com/hashicorp/terraform-provider-nomad/issues/548


jrasell (approved):
LGTM

PR #564, opened 2025-10-28
fix: support string durations in json jobspec

smintz:
In go `time.Duration` is and alias to `int64` and does not implements the `json.Unmarsaler` interface. This means the JSON representation of the jobspec only accepts numeric values for `time.Duration` fields.

In the JSON spec, numbers which are higher than `MaxInt32`, should be represented in strings and should be `"quoted"`. Some tools are very strict about this and does not allow marshaling of numbers which are over `maxInt32` without quoting. Go `json.Unmarshal()` on the other hand, doesn't allow those string notations when decoding the json and returns and `err`.

In this case, generating json jobspecs using these tools results a maximum of ~1 second for any `time.Duration` field.

In this change, I added `jsoniter` as a drop-in replacement for `json.Unmarshal()` which allow registering of decode functions for specific type.

The new decoder maintain compatibility with previously used files using numeric fields, but also accepts `"quoted"` values and even standard `time.ParseDuration()` values (`15s`, `1m`, etc).

### DIFF
--- a/nomad/resource_job_test.go
+++ b/nomad/resource_job_test.go
@@ -1052,7 +1052,7 @@ resource "nomad_job" "test" {
   ],
   "Update": {
 	"Stagger": 30000000000,
-	"MinHealthyTime": "20s",
+	"MinHealthyTime": "5s",
 	"HealthyDeadline": "10000000000",
 	"AutoRevert": true,
 	"MaxParallel": 2


smintz:
@tgross, totally agree about adding `jsoniter`. I would love if I could find a solution using the standard library, but so for I couldn't.
I tried:
- Adding an alias to `time.Duration` within the nomad project. It was too complicated as there are many functions expecting `time.Duration` and using an alias could not just be a drop-in replacement.
- Wrapping `time.Duration`: same issue.
- Introducing `UnmarshalJSON` function within `time.Duration` standard library. Resulted in imports loop and compile error.

My workflow:
I'm using [protoconf](https://github.com/protoconf/protoconf) to manage all of my configs (both software level and infra). I'm wrapping terraform using [protoconf-terraform](https://github.com/protoconf/protoconf) and want to switch from ECS to nomad.
I've generated a compatible protobuf definition file with the nomad jobspec so I can compose jobs from within protoconf.

I made sure my change is not breaking previous workflows and added tests.

tgross:
I'm not going to get a chance to review this today but I should be able to get to it in the next few days. 

But from a high level I think my big worry here is that we're allowing users to submit JSON jobspecs via Terraform that are not valid Nomad JSON jobspecs, because the Nomad API itself doesn't accept those values. But it'd be painful to fix this in the Nomad API not so much from the server side, but because the Go API would need to change the type of fields, so we'd be breaking backwards compatibility (I think any of my colleagues would agree that a hypothetical v2 API would not make this mistake).

tgross:
@smintz something smelled funny to me about the premise of this PR and I went to double-check my memory. I'm not sure where you're getting that the JSON spec supports up to only MaxInt32, but that's not in the Crockford JSON spec nor in the [RFC8259 section on numbers](https://www.rfc-editor.org/rfc/rfc8259#section-6).

The range is explicitly implementation defined, and there's this note because even if you represent the a number with a float it's good up to 2^53:
> Note that when such software is used, numbers that are integers and
> are in the range [-(2**53)+1, (2**53)-1] are interoperable in the
> sense that implementations will agree exactly on their numeric
> values.

What's the implementation you're using that you're getting such a narrow range?

PR #593, opened 2026-03-27
remove secretsMB as it's not a quota field

mismithhisler:
secretsMB should not be set as it's only part of the Quota object because of previous struct sharing.


tgross (approved):
LGTM!

PR #594, opened 2026-03-31
add muxed framework provider scaffolding and migrate acceptance tests to terraform-plugin-testing

vishnuedavarath:
- introduce a protocol v6 mux server combining the current SDKv2 provider with a new framework provider
- reuse SDKv2 provider metadata from the framework provider so the framework resource can use the already-configured Nomad client instead of duplicating provider configuration logic
- align provider schema sensitivity between SDKv2 and framework definitions for secret-bearing fields
- add focused mux tests for merged metadata/schema and an SDK data source through the muxed provider
- migrate SDKv2 testing helper packages to terraform-plugin-testing

[JIRA Ticket](https://hashicorp.atlassian.net/browse/NMD-1297)


internal/mux/mux_test.go
```diff
@@ -0,0 +1,85 @@
...
+
+func TestMuxServer_MetadataAndSchema(t *testing.T) {
+	t.Parallel()
+
+	ctx := t.Context()
+
+	server, err := MuxServer(ctx)
+	if err != nil {
+		t.Fatalf("expected mux server to initialize: %v", err)
+	}
```
pkazmierczak, line 29: We've been slowly migrating over to [shoenig/test](https://github.com/shoenig/test) in the Nomad codebase, and I see there are already some uses in the tf provider. It's not a blocker for me, but maybe this could just be `must.NoError(err, must.Sprintf("expected mux server to initialize: %v", err))`?

pkazmierczak (approved):
LGTM 👍

PR #599, opened 2026-04-06
`nomad_acl_auth_method` / `nomad_acl_binding_rule` Update to framework and add write only properties

danquack:
ACL Auth Method has secrets stored in state. With the muxing/upgrade to framework users can now specify these properties as WriteOnly. In order to implement this the resource itself had to be updated to framework, and then the resource attributes could be added.

The following attributes were added to the resource:
* `config.oidc_client_secret_wo`
* `config.oidc_client_secret_wo_version`
* `private_key.pem_key_wo`
* `private_key.pem_key_wo_version`

WO Version is optional when secrets are set using [private state](https://dev.to/drewmullen/writing-terraform-resources-with-write-only-parameters-570j) to determine the value.

All logs also have been migrated to tflog

### Tests
```
# TF_ACC=1 go test -count=1 -run TestResourceACL  ./...
?       github.com/hashicorp/terraform-provider-nomad   [no test files]
?       github.com/hashicorp/terraform-provider-nomad/internal/framework/provider       [no test files]
ok      github.com/hashicorp/terraform-provider-nomad/internal/framework/provider/acl   4.546s
?       github.com/hashicorp/terraform-provider-nomad/internal/framework/provider/testutil      [no test files]
ok      github.com/hashicorp/terraform-provider-nomad/internal/mux      0.769s [no tests to run]
ok      github.com/hashicorp/terraform-provider-nomad/nomad     9.438s
?       github.com/hashicorp/terraform-provider-nomad/nomad/helper      [no test files]
?       github.com/hashicorp/terraform-provider-nomad/nomad/helper/pointer      [no test files]
➜  terraform-provider-nomad git:(feat/add-writeonly) TF_ACC=1 go test -count=1 -run '^TestResourceACL'  ./...
?       github.com/hashicorp/terraform-provider-nomad   [no test files]
?       github.com/hashicorp/terraform-provider-nomad/internal/framework/provider       [no test files]
ok      github.com/hashicorp/terraform-provider-nomad/internal/framework/provider/acl   4.482s
?       github.com/hashicorp/terraform-provider-nomad/internal/framework/provider/testutil      [no test files]
ok      github.com/hashicorp/terraform-provider-nomad/internal/mux      0.348s [no tests to run]
ok      github.com/hashicorp/terraform-provider-nomad/nomad     9.170s
?       github.com/hashicorp/terraform-provider-nomad/nomad/helper      [no test files]
?       github.com/hashicorp/terraform-provider-nomad/nomad/helper/pointer      [no test files]
➜  terraform-provider-nomad git:(feat/add-writeonly) TF_ACC=1 go test -v -count=1 -run '^TestResourceACL'  ./...
?       github.com/hashicorp/terraform-provider-nomad   [no test files]
?       github.com/hashicorp/terraform-provider-nomad/internal/framework/provider       [no test files]
=== RUN   TestResourceACLBindingRule
--- PASS: TestResourceACLBindingRule (0.92s)
=== RUN   TestResourceACLManagementBindingRule
--- PASS: TestResourceACLManagementBindingRule (0.56s)
=== RUN   TestResourceACLManagementBindingRule_upgradeToFramework
--- PASS: TestResourceACLManagementBindingRule_upgradeToFramework (2.39s)
PASS
ok      github.com/hashicorp/terraform-provider-nomad/internal/framework/provider/acl   4.220s
ok      github.com/hashicorp/terraform-provider-nomad/internal/mux      0.628s [no tests to run]
=== RUN   TestResourceACLPolicy_import
--- PASS: TestResourceACLPolicy_import (0.75s)
=== RUN   TestResourceACLPolicy_basic
--- PASS: TestResourceACLPolicy_basic (0.53s)
=== RUN   TestResourceACLPolicy_refresh
--- PASS: TestResourceACLPolicy_refresh (0.82s)
=== RUN   TestResourceACLPolicy_nameChange
--- PASS: TestResourceACLPolicy_nameChange (0.91s)
=== RUN   TestResourceACLPolicy_update
--- PASS: TestResourceACLPolicy_update (0.87s)
=== RUN   TestResourceACLPolicy_namespaceOnlyJobACL
--- PASS: TestResourceACLPolicy_namespaceOnlyJobACL (0.50s)
=== RUN   TestResourceACLRole
--- PASS: TestResourceACLRole (0.88s)
=== RUN   TestResourceACLToken_import
--- PASS: TestResourceACLToken_import (0.68s)
=== RUN   TestResourceACLToken_basic
--- PASS: TestResourceACLToken_basic (0.50s)
=== RUN   TestResourceACLToken_update
--- PASS: TestResourceACLToken_update (0.84s)
=== RUN   TestResourceACLToken_Expiration
--- PASS: TestResourceACLToken_Expiration (0.52s)
=== RUN   TestResourceACLToken_RoleLink
--- PASS: TestResourceACLToken_RoleLink (0.56s)
PASS
ok      github.com/hashicorp/terraform-provider-nomad/nomad     9.270s
?       github.com/hashicorp/terraform-provider-nomad/nomad/helper      [no test files]
?       github.com/hashicorp/terraform-provider-nomad/nomad/helper/pointer      [no test files]
```

## Maintainer Notes
-  [x] Docs Update?
```
 make website
make[1]: *** No rule to make target `website-provider'.  Stop.
```


danquack:
@vishnuedavarath @pkazmierczak  I migrated `nomad_acl_binding_rule` as well, so this should be good to kick the tests again

internal/framework/provider/acl/resource_acl_auth_method.go
```diff
@@ -0,0 +1,1207 @@
...
+						Description: "The OAuth Client Secret configured with the OIDC provider. This value is write-only and will not be stored in Terraform state.",
+						Validators: []validator.String{
+							stringvalidator.ConflictsWith(path.MatchRelative().AtParent().AtName("oidc_client_secret")),
+						},
+					},
+					"oidc_client_secret_wo_version": schema.Int64Attribute{
+						Optional:    true,
+						Computed:    true,
+						Description: "Version counter for oidc_client_secret_wo. Increments automatically when the write-only secret changes, or set manually to trigger an update.",
+					},
```
vishnuedavarath, line 200: The write-only version fields should be constrained to the write-only inputs only. I think `oidc_client_secret_wo_version` should require `oidc_client_secret_wo` when explicitly configured, and conflict with `oidc_client_secret`, since the version marker is only meaningful for the write-only secret path.

internal/framework/provider/acl/resource_acl_auth_method.go
```diff
@@ -0,0 +1,1207 @@
...
+										Description: "RSA private key PEM to use to sign the JWT. This value is write-only and will not be stored in Terraform state.",
+										Validators: []validator.String{
+											stringvalidator.ConflictsWith(path.MatchRelative().AtParent().AtName("pem_key")),
+										},
+									},
+									"pem_key_wo_version": schema.Int64Attribute{
+										Optional:    true,
+										Computed:    true,
+										Description: "Version counter for pem_key_wo. Increments automatically when the write-only key changes, or set manually to trigger an update.",
+									},
```
vishnuedavarath, line 331: Similar to `oidc_client_secret_wo_version`

internal/framework/provider/acl/resource_acl_auth_method.go
vishnuedavarath: After looking at the plan logic, I don’t think `oidc_client_secret_wo_version` needs to require with `oidc_client_secret_wo`, since it’s auto-populated for that path. I do still think it should conflict with `oidc_client_secret`, because the version marker only applies to the write-only flow.

vishnuedavarath (changes requested):
I think this could also use docs and a changelog entry, since it adds user-visible write-only behavior.

danquack:
> I think this could also use docs and a changelog entry, since it adds user-visible write-only behavior.

The make website-provider doesn't work for me. I don't have the binary, so not clear on how to get that.

internal/framework/provider/acl/resource_acl_auth_method.go
```diff
@@ -0,0 +1,1213 @@
...
+						Validators: []validator.String{
+							stringvalidator.ConflictsWith(path.MatchRelative().AtParent().AtName("oidc_client_secret")),
+						},
+					},
+					"oidc_client_secret_wo_version": schema.Int64Attribute{
+						Optional:    true,
+						Computed:    true,
+						Description: "Version counter for oidc_client_secret_wo. Increments automatically when the write-only secret changes, or set manually to trigger an update.",
+						Validators: []validator.String{
+							stringvalidator.ConflictsWith(path.MatchRelative().AtParent().AtName("oidc_client_secret")),
```
vishnuedavarath: `oidc_client_secret_wo_version` and `pem_key_wo_version` are defined as `schema.Int64Attribute`, but their validators are typed as []validator.String using `stringvalidator.ConflictsWith(...)`. That causes the build failure because framework validators are attribute-type specific. These need [`]validator.Int64` validators instead.

vishnuedavarath (changes requested):

vishnuedavarath:
> > I think this could also use docs and a changelog entry, since it adds user-visible write-only behavior.
> 
> The make website-provider doesn't work for me. I don't have the binary, so not clear on how to get that.

I tried a bit to get it working, but it looks like legacy tooling and I didn’t spend more time trying to fix it. I ended up updating the docs manually. We can use https://registry.terraform.io/tools/doc-preview for previewing if needed.

internal/framework/provider/acl/resource_acl_auth_method.go
danquack: a classic copy pasta issue!

danquack:
looks like your [fix](https://github.com/hashicorp/terraform-provider-nomad/pull/613) [worked](https://github.com/danquack/terraform-provider-nomad/actions/runs/24675174233/job/72156939430) @vishnuedavarath!

vishnuedavarath (approved):
LGTM!

internal/framework/provider/acl/resource_acl_binding_rule.go
```diff
@@ -0,0 +1,302 @@
...
+
+	var old v0State
+	if err := json.Unmarshal(req.RawState.JSON, &old); err != nil {
+		resp.Diagnostics.AddError("Failed to parse prior state for upgrade", err.Error())
+		return
+	}
+
+	upgraded := aclBindingRuleModel{
+		ID:          types.StringValue(old.ID),
+		Description: types.StringValue(old.Description),
```
pkazmierczak: I think there's a tiny bug in here. 

The SDKv2 schema for the binding rule has `description` and `selector` as plain `Optional` strings; no `Default`, no `Computed`. In SDKv2, an unset Optional string is persisted as `""` in the state JSON. So a user who never wrote `description` in their HCL would have `"description": ""` in their v0 state.

After upgrade, the state holds `types.StringValue("")`. In the v1 schema: `description` is `Optional` only (not Computed). For a non-Computed attribute, the Framework plans the config value. If the user's HCL has no `description` line, config is `null`. The Framework sees `state = ""` ≠ `plan = null` and shows a `~ description: "" -> null` diff. The apply sends `""` to Nomad (same as before), state becomes null, and it stabilizes. Not catastrophic, but it's a surprise for every user who didn't have `description` or `selector` in their config.

```suggestion
		Description: optionalStringFromAPI(old.Description),
```

(easy fix, and the same applies to `Selector` field)

internal/framework/provider/acl/resource_acl_binding_rule.go
```diff
@@ -0,0 +1,302 @@
...
+		AuthMethod:  types.StringValue(old.AuthMethod),
+		Selector:    types.StringValue(old.Selector),
```
pkazmierczak:
```suggestion
		Selector:    optionalStringFromAPI(old.Selector),
```

internal/framework/provider/acl/resource_acl_auth_method_test.go
```diff
@@ -0,0 +1,672 @@
...
+		CheckDestroy: testACLAuthMethodDestroyed(t, methodName),
+		Steps: []resource.TestStep{
+			{
+				ExternalProviders: map[string]resource.ExternalProvider{
+					"nomad": {
+						Source:            "hashicorp/nomad",
+						VersionConstraint: "2.5.2",
+					},
+				},
+				Config: testACLAuthMethodConfigOIDCWithMaxTTL(methodName, "someclientsecret", "http://localhost:4649/oidc/callback", false, "10m0s"),
```
pkazmierczak: This test is good, but it only provisions a plain OIDC config. There's no test upgrading a resource that has the `oidc_client_assertion` + `private_key` nested block in its v0 state. 

It's a nice-to-have, `TestACLAuthMethod_upgradeToFramework` logic looks correct to me. But it'd be nice to have an acceptance test for the code path I mention (lines 830--852 in the resource).

pkazmierczak (commented):
Wow, great effort @danquack! I left 2 minor comments that I'd like you to look at, but otherwise this looks good to me.

internal/framework/provider/acl/resource_acl_auth_method_test.go
danquack: Updated to `clientAssertionResourcesHCL` which `private_key` in the last commit. I think that should resolve it?
```
TF_ACC=1 go test $(go list ./...) -v -run=TestACLAuthMethod_upgradeToFramework -timeout 120m -count=1
?   	github.com/hashicorp/terraform-provider-nomad	[no test files]
?   	github.com/hashicorp/terraform-provider-nomad/internal/framework/provider	[no test files]
=== RUN   TestACLAuthMethod_upgradeToFramework
--- PASS: TestACLAuthMethod_upgradeToFramework (10.65s)
PASS
```

pkazmierczak (approved):
LGTM! Amazing work! 🎉

PR #600, opened 2026-04-14
resource/nomad_variable: require items_wo_version with items_wo

vishnuedavarath:
### Description

Previously, `items_wo_version` required `items_wo`, but `items_wo` could still be set without `items_wo_version`. That was inconsistent with how the resource actually applies write-only updates, since write-only payload changes are driven by `items_wo_version`.

This update makes the contract explicit and consistent:

- `items` and `items_wo` remain mutually exclusive
- `items_wo` now requires `items_wo_version`
- `items_wo_version` continues to conflict with `items`

The change also updates validation tests to cover the invalid combinations and clarifies the schema descriptions/runtime messaging for the write-only flow.

### DIFF
--- a/nomad/resource_variable.go
+++ b/nomad/resource_variable.go
@@ -63,19 +63,19 @@ func resourceVariable() *schema.Resource {
 				ExactlyOneOf: []string{"items", "items_wo"},
 			},
 			"items_wo": {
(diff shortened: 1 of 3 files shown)


pkazmierczak (approved):
👍

PR #602, opened 2026-04-16
update terraform registry protocol version

vishnuedavarath:
### Description

This PR updates the Terraform Registry manifest from protocol v5 to v6 to align with the provider’s current main and mux-based implementation. No functional provider behavior changes are intended beyond the protocol/version metadata update.

### DIFF
--- a/terraform-registry-manifest.json
+++ b/terraform-registry-manifest.json
@@ -1,6 +1,6 @@
 {
     "version": 1,
     "metadata": {
-        "protocol_versions": ["5.0"]
+        "protocol_versions": ["6.0"]
     }
 }


terraform-registry-manifest.json
```diff
@@ -1,6 +1,6 @@
 {
     "version": 1,
     "metadata": {
-        "protocol_versions": ["5.0"]
+        "protocol_versions": ["6.0"]
```
gulducat, line 4: What prompted this change? Based on the documentation here
* https://developer.hashicorp.com/terraform/registry/providers/publishing

we should be using 5.0, because this provider is built with TF Plugin SDK v2

I.e. this package

```
$ grep plugin-sdk go.mod
        github.com/hashicorp/terraform-plugin-sdk/v2 v2.40.0
```

vishnuedavarath: We're actually using a mux server now. It's protocol v6.
https://github.com/hashicorp/terraform-provider-nomad/pull/594

gulducat (approved):
Thanks for the clarification, LGTM!

PR #604, opened 2026-04-16
omit x5 suffix for the bin name (#603)

vishnuedavarath:


gulducat (approved):
sure hope it doesn't try to auto-delete `main` after merge 😋

PR #605, opened 2026-04-16
prepare next release 2.6.0

vishnuedavarath:
### Description

Prepare for next release


.github/workflows/build.yml
gulducat, line 140: We could have done some `git` procedure better here, as this is already in `main`, but the code is identical, so no mergeflict, so no worries.

gulducat (approved):
🚀

PR #607, opened 2026-04-16
Update docs to tfplugindocs

danquack:
### Description
Not sure if this is desired per https://github.com/hashicorp/terraform-provider-nomad/pull/599#issuecomment-4251870218. But updated the docs to use [tfplugindocs](https://github.com/hashicorp/terraform-plugin-docs) which will make it so manual doc updates aren't needed.

Also added some GitHub actions to:
- Validate Docs work with current binary
- Validate user ran make website


internal/framework/provider/provider.go
```diff
@@ -41,7 +41,7 @@ func (p *NomadProvider) Schema(_ context.Context, _ frameworkprovider.SchemaRequ
 	resp.Schema = providerschema.Schema{
 		Attributes: map[string]providerschema.Attribute{
 			"address": providerschema.StringAttribute{
-				Optional:    true,
+				Required:    true,
```
danquack: This was causing build issues because it mismatches with sdk. https://github.com/hashicorp/terraform-provider-nomad/blob/main/nomad/provider.go#L27-L31

PR #615, opened 2026-04-25
nomad_acl_token: fix perpetual ForceNew on expiration_ttl via DiffSuppressFunc

### LINKED ISSUE #614: [Bug] nomad_acl_token: expiration_ttl causes perpetual forced replacement due to time.Duration normalization
### Description

When `expiration_ttl` is set to a duration string without trailing zero components (e.g. `"1h"`, `"30m"`, `"24h"`), the `nomad_acl_token` resource is destroyed and recreated on every subsequent `terraform plan`, even with no configuration changes.

### Root Cause

`resourceACLTokenRead()` writes the expiration TTL back to state using Go's `time.Duration.String()`:

```go
// resource_acl_token.go:193
d.Set("expiration_ttl", token.ExpirationTTL.String())
```

`time.Duration.String()` always appends zero-value sub-components:
- `"1h"` → `"1h0m0s"`
- `"30m"` → `"30m0s"`
- `"90s"` → `"1m30s"`

Since `expiration_ttl` has `ForceNew: true` and no `DiffSuppressFunc`, Terraform interprets `"1h" ≠ "1h0m0s"` as a configuration change requiring destroy and recreate. This rotates the `secret_id` on every apply, silently invalidating any downstream services that depend on the token.

### Steps to Reproduce

```hcl
resource "nomad_acl_token" "example" {
  name           = "example"
  type           = "management"
  expiration_ttl = "1h"
}
```

```
$ terraform apply    # succeeds
$ terraform plan     # shows forced replacement with no config changes

  # nomad_acl_token.example must be replaced
  -/+ resource "nomad_acl_token" "example" {
      ~ expiration_ttl = "1h0m0s" -> "1h" # forces replacement
      ~ secret_id      = (sensitive value)
    }

Plan: 1 to add, 0 to change, 1 to destroy.
```

Applying again rotates `secret_id`. The cycle repeats indefinitely.

### Environment

- Provider: `hashicorp/nomad` **v2.6.1** (latest, verified on main branch as of 2026-04-25)
- Nomad server: **v2.0.0** (latest)
- Terraform: **v1.5.7**
- OS: darwin/arm64

### Expected Behavior

`terraform plan` reports **No changes** after a successful apply with no configuration changes.

### Fix

Add a `DiffSuppressFunc` that compares the parsed duration values, identical to the existing fix already applied to `max_token_ttl` in `resource_acl_auth_method.go`:

```go
DiffSuppressFunc: func(k, oldValue, newValue string, d *schema.ResourceData) bool {
    o, _ := time.ParseDuration(oldValue)
    n, _ := time.ParseDuration(newValue)
    return o == n
},
```

A pull request with this fix is attached.

YuechengHao626:
## Summary

Fixes #614

`expiration_ttl` on `nomad_acl_token` caused a perpetual destroy-and-recreate cycle when set to durations like `"1h"` or `"30m"`. `Read()` writes back the value via `token.ExpirationTTL.String()`, which Go normalizes to `"1h0m0s"` — a string that does not match the original config value. With `ForceNew: true` and no `DiffSuppressFunc`, every `terraform plan` schedules a replacement, silently rotating `secret_id` and breaking downstream services.

## Change

Added a `DiffSuppressFunc` to `expiration_ttl` that parses both old and new values with `time.ParseDuration` and suppresses the diff when they represent the same duration.

This is the same pattern already used in this repo for `max_token_ttl` in `resource_acl_auth_method.go`.

## Testing

Verified locally with provider built from this branch against Nomad v2.0.0:

```
# Before fix
terraform plan → expiration_ttl = "1h0m0s" -> "1h" # forces replacement  (exit code 2)

# After fix
terraform plan → No changes.  (exit code 0)
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## UNRELEASED
 
+BUG FIXES:
+* resource/nomad_acl_token: Fixed perpetual destroy-and-recreate cycle when `expiration_ttl` is set to a duration like `"1h"` or `"30m"`. `Read()` normalizes the value via Go's `time.Duration.String()` (e.g. `"1h"` → `"1h0m0s"`), causing a diff with `ForceNew: true`. Added a `DiffSuppressFunc` that compares durations by value using `time.ParseDuration`. ([#615](https://github.com/hashicorp/terraform-provider-nomad/pull/615))
+
 ## 2.6.1 (April 20, 2026)
 
 BUG FIXES:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 ## UNRELEASED
 
 BUG FIXES:
-* resource/nomad_acl_token: Fixed perpetual destroy-and-recreate cycle when `expiration_ttl` is set to a duration like `"1h"` or `"30m"`. `Read()` normalizes the value via Go's `time.Duration.String()` (e.g. `"1h"` → `"1h0m0s"`), causing a diff with `ForceNew: true`. Added a `DiffSuppressFunc` that compares durations by value using `time.ParseDuration`. ([#615](https://github.com/hashicorp/terraform-provider-nomad/pull/615))
+* resource/nomad_acl_token: Fixed perpetual destroy-and-recreate cycle when `expiration_ttl` is set to a duration like `"1h"` or `"30m"`. ([#615](https://github.com/hashicorp/terraform-provider-nomad/pull/615))
 
 ## 2.6.1 (April 20, 2026)
 
--- a/nomad/resource_acl_token.go
+++ b/nomad/resource_acl_token.go
@@ -91,6 +91,11 @@ func resourceACLToken() *schema.Resource {
 				Default:     "0s",
 				ForceNew:    true,
 				Type:        schema.TypeString,
+				DiffSuppressFunc: func(k, oldValue, newValue string, d *schema.ResourceData) bool {
+					o, _ := time.ParseDuration(oldValue)
+					n, _ := time.ParseDuration(newValue)
+					return o == n
+				},
 			},
 			"expiration_time": {
 				Description: "The point after which a token is considered expired and eligible for destruction.",
--- a/nomad/resource_acl_token.go
+++ b/nomad/resource_acl_token.go
@@ -92,8 +92,11 @@ func resourceACLToken() *schema.Resource {
 				ForceNew:    true,
 				Type:        schema.TypeString,
 				DiffSuppressFunc: func(k, oldValue, newValue string, d *schema.ResourceData) bool {
-					o, _ := time.ParseDuration(oldValue)
-					n, _ := time.ParseDuration(newValue)
+					o, err1 := time.ParseDuration(oldValue)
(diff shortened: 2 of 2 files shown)


nomad/resource_acl_token.go
```diff
@@ -91,6 +91,11 @@ func resourceACLToken() *schema.Resource {
 				Default:     "0s",
 				ForceNew:    true,
 				Type:        schema.TypeString,
+				DiffSuppressFunc: func(k, oldValue, newValue string, d *schema.ResourceData) bool {
+					o, _ := time.ParseDuration(oldValue)
+					n, _ := time.ParseDuration(newValue)
```
tgross: Can we return `false` on error here? That'd give us a diff and alert the user that something's wrong.

tgross (changes requested):
Hi @YuechengHao626! Thanks for the PR. I've left a comment about error handling. Also, can you add a changelog entry?

nomad/resource_acl_token.go
YuechengHao626: Thanks for the suggestion! Updated in the latest commit.

YuechengHao626:
> Hi @YuechengHao626! Thanks for the PR. I've left a comment about error handling. Also, can you add a changelog entry?

Thanks for the review! Done in the latest commit.

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 ## UNRELEASED
 
+BUG FIXES:
+* resource/nomad_acl_token: Fixed perpetual destroy-and-recreate cycle when `expiration_ttl` is set to a duration like `"1h"` or `"30m"`. `Read()` normalizes the value via Go's `time.Duration.String()` (e.g. `"1h"` → `"1h0m0s"`), causing a diff with `ForceNew: true`. Added a `DiffSuppressFunc` that compares durations by value using `time.ParseDuration`. ([#615](https://github.com/hashicorp/terraform-provider-nomad/pull/615))
```
tgross: This is a changelog and not a commit message, so let's whittle this down a bit to the user-facing behavior:

```suggestion
* resource/nomad_acl_token: Fixed perpetual destroy-and-recreate cycle when `expiration_ttl` is set to a duration like `"1h"` or `"30m"`. ([#615](https://github.com/hashicorp/terraform-provider-nomad/pull/615))
```

YuechengHao626: > This is a changelog and not a commit message, so let's whittle this down a bit to the user-facing behavior:

I've addressed your comments.

tgross (approved):

tgross:
@YuechengHao626 I didn't realize this until I hit squash merge, but in the future you should clearly identify in your PR that you've used a LLM to do the work. Signing the CLA is in this case is _really_ dubious.

YuechengHao626:
> @YuechengHao626 I didn't realize this until I hit squash merge, but in the future you should clearly identify in your PR that you've used a LLM to do the work. Signing the CLA is in this case is _really_ dubious.

Thanks for your understanding.  Yes, I used an LLM to assist with writing the code, but the bug analysis, root cause identification, fix design, and code review were all done by me. I'll make sure to disclose that upfront in the future PRs.

PR #621, opened 2026-05-10
fix: nomad_job honors Ctrl+C during monitorDeployment by adopting con…

### LINKED ISSUE #620: nomad_job: Ctrl+C / context cancel ignored during monitorDeployment (detach=false)
## Summary

`terraform apply` cannot be interrupted while a `nomad_job` resource with `detach = false` is waiting for its deployment. After Ctrl+C, Terraform prints `Stopping operation...` but the apply blocks until either the Nomad deployment reaches a terminal state (e.g. `progress_deadline` elapses) or the provider's internal `Timeouts.Create` fires - which can be many minutes.

This is the same class of bug as #367 (Ctrl+C ignored during volume deletion), which was fixed by adopting SDK v2's context-aware function signatures. The fix was applied to volume operations but never extended to `nomad_job`.

## Reproduction

```hcl
resource "nomad_job" "example" {
  jobspec = file("${path.module}/slow-to-converge.nomad.hcl")
  detach  = false
  timeouts {
    create = "30m"
  }
}
```

Pick a jobspec whose deployment takes a while to converge (long `min_healthy_time`, slow startup, etc.). Run `terraform apply`, and Ctrl+C during the create. Terraform's interrupt handler prints `Stopping operation...` but the resource keeps reporting `Still creating... [Xm elapsed]` until the Nomad deployment reaches `successful`/`failed`/`cancelled` or the resource's own timeout expires.

A second Ctrl+C force-kills Terraform mid-write and corrupts state, which is exactly what the SDK v2 context plumbing is supposed to prevent.

## Root cause

`nomad_job` is wired to the legacy non-context-aware CRUD functions ([resource_job.go#L28-L31](https://github.com/hashicorp/terraform-provider-nomad/blob/main/nomad/resource_job.go#L28-L31)):

```go
Create: resourceJobRegister,
Update: resourceJobRegister,
Delete: resourceJobDeregister,
Read:   resourceJobRead,
```

`resourceJobRegister` has the legacy signature ([resource_job.go#L505](https://github.com/hashicorp/terraform-provider-nomad/blob/main/nomad/resource_job.go#L505)):

```go
func resourceJobRegister(d *schema.ResourceData, meta interface{}) error {
```

so it never receives the cancellable `context.Context` Terraform passes to providers on signal. As a result, `monitorDeployment` falls back to `context.Background()` ([resource_job.go#L606,626](https://github.com/hashicorp/terraform-provider-nomad/blob/main/nomad/resource_job.go#L606)):

```go
state, err := stateConf.WaitForStateContext(context.Background())
```

`WaitForStateContext(context.Background())` cannot observe Terraform's cancel; the wait runs to its own internal completion regardless of the user's interrupt. `resourceJobDeregister` has the same shape and presumably the same problem on destroy.

## Proposed fix

Mirror what was done for volumes after #367.

## Provider / Terraform versions

- Provider: `hashicorp/nomad` v2.6.1 (also reproduces against `main` as of writing)
- Terraform: 1.14.7

maciej-lech:
…text-aware CRUD

### Description

Fixes #620 

See #620 

### DIFF
--- a/nomad/resource_job.go
+++ b/nomad/resource_job.go
@@ -28,8 +28,8 @@ func resourceJob() *schema.Resource {
 	return &schema.Resource{
 		CreateContext: resourceJobRegister,
 		UpdateContext: resourceJobRegister,
-		Delete:        resourceJobDeregister,
-		Read:          resourceJobRead,
+		DeleteContext: resourceJobDeregister,
+		ReadContext:   resourceJobRead,
 
 		CustomizeDiff: resourceJobCustomizeDiff,
 
@@ -588,7 +588,7 @@ func resourceJobRegister(ctx context.Context, d *schema.ResourceData, meta inter
 		}
 	}
 
-	return diag.FromErr(resourceJobRead(d, meta)) // populate other computed attributes
+	return resourceJobRead(ctx, d, meta) // populate other computed attributes
 }
 
 // monitorDeployment monitors the evalution(s) from a job create/update and,
@@ -701,7 +701,7 @@ func deploymentStateRefreshFunc(client *api.Client, namespace string, deployment
 	}
 }
 
-func resourceJobDeregister(d *schema.ResourceData, meta interface{}) error {
+func resourceJobDeregister(_ context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
 	providerConfig := meta.(ProviderConfig)
 	client := providerConfig.client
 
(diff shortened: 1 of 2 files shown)


nomad/resource_job.go
```diff
@@ -25,10 +26,10 @@ import (
...
 func resourceJob() *schema.Resource {
 	return &schema.Resource{
-		Create: resourceJobRegister,
-		Update: resourceJobRegister,
-		Delete: resourceJobDeregister,
-		Read:   resourceJobRead,
+		CreateContext: resourceJobRegister,
+		UpdateContext: resourceJobRegister,
+		DeleteContext: resourceJobDeregister,
+		Read:          resourceJobRead,
```
tehut: I noticed you disregarded the context object for DeleteContext (presumably because it's only called in destroys). Is there a reason you didn't do the same thing for Read?

maciej-lech: I blindly followed #371, but here `Delete` doesn't need context. Let me change `Delete` to context-less version.

maciej-lech: Now it follows volume precedent. If you prefer context version for all of them, and just discard them if not needed, just say a word ;)

tehut: Yeah, I think it's probably better to have them all be consistent.

maciej-lech: Done

tehut (approved):
👍

PR #625, opened 2026-05-21
fix(datasource/nomad_variable): fix panic when items_wo_version is not in schema

### LINKED ISSUE #623: data "nomad_variable" Read panics: "interface conversion: interface {} is nil, not int" (v2.6.0+)
Hi there,

### Terraform Version

```
Terraform v1.14.7
on linux_amd64
+ provider registry.terraform.io/hashicorp/nomad v2.6.1
```

Also reproduces with v2.6.0.

### Nomad Version

```
Nomad v1.10.5
```

### Provider Configuration

```hcl
provider "nomad" {}
```

### Environment Variables

None Nomad-specific (defaults).

### Affected Resource(s)

- `data "nomad_variable"`

```hcl
resource "nomad_variable" "test" {
  path  = "example/test"
  items = { secret = "abc" }
}

data "nomad_variable" "test" {
  path       = "example/test"
  depends_on = [nomad_variable.test]
}

output "secret" {
  value     = data.nomad_variable.test.items["secret"]
  sensitive = true
}
```

### Debug Output

Not needed - root cause is identifiable from the panic stack and the provider source (see below).

### Panic Output

```
panic: interface conversion: interface {} is nil, not int

goroutine 34 [running]:
github.com/hashicorp/terraform-provider-nomad/nomad.resourceVariableRead(...)
        github.com/hashicorp/terraform-provider-nomad/nomad/resource_variable.go:155 +0x35e
github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema.(*Resource).read(...)
        github.com/hashicorp/terraform-plugin-sdk/v2@v2.40.0/helper/schema/resource.go:854 +0x15f
github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema.(*Resource).ReadDataApply(...)
        github.com/hashicorp/terraform-plugin-sdk/v2@v2.40.0/helper/schema/resource.go:1088 +0x13e
github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema.(*GRPCProviderServer).ReadDataSource(...)
        github.com/hashicorp/terraform-plugin-sdk/v2@v2.40.0/helper/schema/grpc_provider.go:1994 +0x6af
...
```

### Expected Behavior

`data "nomad_variable"` reads the variable and returns its `items` map.

### Actual Behavior

The provider plugin panics. The crash happens at `nomad/resource_variable.go:155`, where the shared Read function dereferences a schema attribute that the data source does not declare:

```go
// resource_variable.go:155
if d.Get("items_wo_version").(int) > 0 {
    return d.Set("items", nil)
}
```

`data "nomad_variable"` is wired to this same Read function (`data_source_variable.go:13` - `Read: resourceVariableRead`), but its schema (`data_source_variable.go:16-34`) only declares `path`, `namespace`, and `items` - **no `items_wo_version`**. So `d.Get("items_wo_version")` returns `nil`, and the `.(int)` type assertion panics.

This is a regression introduced by #598 / #600 (merged 2026-04-13 / 2026-04-14, shipped in v2.6.0), which added `items_wo_version` to the resource schema and the shared Read function without updating the data-source schema or splitting the Read paths.

### Steps to Reproduce

1. `terraform apply` the snippet above against any reachable Nomad cluster running v2.6.0 or v2.6.1 of the provider.
2. The resource is created.
3. On the next refresh / apply, the data source's Read fires and the provider panics with the trace above.

### Important Factoids

None - reproduces on a standard Nomad install.

### Suggested fix

Two options:

1. Make the assertion safe in the shared Read function:

   ```go
   if v, ok := d.GetOk("items_wo_version"); ok && v.(int) > 0 {
       return d.Set("items", nil)
   }
   ```

2. Split the data-source Read out into its own function. The data source has no `items_wo_version` semantics (it always wants `items` back), so a dedicated `dataSourceVariableRead` reads cleaner and removes a shared-mutable-state hazard.

Note: `data "nomad_variable"` is already marked deprecated in v2.6 in favor of `ephemeral "nomad_variable"`. Even so, it should not panic during its deprecation window.

### References

- #598 - add ephemeral nomad_variable and write-only variable support (merged 2026-04-13)
- #600 - resource/nomad_variable: require items_wo_version with items_wo (merged 2026-04-14)
- v2.6.0 release (2026-04-16) - first release containing the regression

vishnuedavarath:
### Description

The shared `resourceVariableRead` function panics with a nil type assertion when called from the data source, which does not declare `items_wo_version` in its schema. Use `d.GetOk` to safely check for the attribute.

Add data source acceptance test to prevent future regressions.

Fixes #623 

[JIRA Ticket](https://hashicorp.atlassian.net/browse/NMD-1491)

1. Reproduce the panic (before fix) :- apply the following config with provider v2.6.0 or v2.6.1:
```terraform
resource "nomad_variable" "test" {
  path  = "example/test"
  items = { secret = "abc" }
}

data "nomad_variable" "test" {
  path       = "example/test"
  depends_on = [nomad_variable.test]
}

output "secret" {
  value     = data.nomad_variable.test.items["secret"]
  sensitive = true
}
```
Expected panic:
```
panic: interface conversion: interface {} is nil, not int
```

2. Verify the fix :- rebuild the provider from this branch and re-run the same config. The data source should read successfully and return items without panic.
3. Run acceptance tests:
```bash
TF_ACC=1 go test ./nomad -run TestDataSourceVariable -count=1 -v
TF_ACC=1 go test ./nomad -run TestResourceVariable -count=1 -v
```
Both should pass :- confirming the data source works and the resource's `items_wo_version` behavior is unaffected.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## UNRELEASED
 
 BUG FIXES:
+* data source/nomad_variable: Fix panic when reading a variable due to `items_wo_version` not being in the data source schema. ([#625](https://github.com/hashicorp/terraform-provider-nomad/pull/625))
 * resource/nomad_acl_token: Fixed perpetual destroy-and-recreate cycle when `expiration_ttl` is set to a duration like `"1h"` or `"30m"`. ([#615](https://github.com/hashicorp/terraform-provider-nomad/pull/615))
 
 ## 2.6.1 (April 20, 2026)
--- a/nomad/data_source_variable_test.go
+++ b/nomad/data_source_variable_test.go
@@ -1,0 +1,47 @@
+// Copyright IBM Corp. 2016, 2025
+// SPDX-License-Identifier: MPL-2.0
+
+package nomad
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/acctest"
+	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+)
+
+func TestDataSourceVariable_basic(t *testing.T) {
+	path := acctest.RandomWithPrefix("tf-nomad-test")
+
+	resource.Test(t, resource.TestCase{
+		Providers: testProviders,
+		PreCheck:  func() { testAccPreCheck(t); testCheckMinVersion(t, "1.4.0") },
+		Steps: []resource.TestStep{
+			{
+				Config: testDataSourceVariable_config(path),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("data.nomad_variable.test", "path", path),
+					resource.TestCheckResourceAttr("data.nomad_variable.test", "items.test_key", "test_value"),
+				),
+			},
+		},
+	})
+}
+
+func testDataSourceVariable_config(path string) string {
+	return fmt.Sprintf(`
+resource "nomad_variable" "test" {
+  path = "%s"
+
+  items = {
+    test_key = "test_value"
+  }
+}
+
+data "nomad_variable" "test" {
+  path       = "%s"
+  depends_on = [nomad_variable.test]
+}
+`, path, path)
+}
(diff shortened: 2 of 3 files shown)


tehut (approved):
🚢

PR #628, opened 2026-05-29
feat(csi-volume): migrate CSI volume resources to framework provider with write-only secrets

vishnuedavarath:
### Description

Migrate nomad_csi_volume and nomad_csi_volume_registration from the legacy SDK to the Terraform plugin framework. Introduce a secrets_wo write-only attribute to prevent sensitive values from being persisted in state, while preserving the existing secrets attribute (deprecated) for backward compatibility.

#### Prerequisites

1. Start a Nomad dev agent with the hostpath CSI plugin

2. Build and install the provider locally:

```bash
make build
```

### Terraform Configuration

```hcl
terraform {
  required_version = ">= 1.11"

  required_providers {
    nomad = {
      source = "hashicorp/nomad"
    }
  }
}

provider "nomad" {
}

resource "nomad_csi_volume" "example" {
  plugin_id    = "hostpath-plugin"
  volume_id    = "test-volume"
  name         = "test-volume"
  capacity_min = "10GiB"
  capacity_max = "20GiB"

  capability {
  }

  mount_options {
    fs_type = "ext4"
  }

  secrets_wo = jsonencode({
    "secret-key-1" = "secret-val-1"
    "secret-key-2" = "secret-val-2"
  })

  topology_request {
    required {
      topology {
        segments = {
          "topology.hostpath.csi/node" = "node-0"
        }
      }
    }
  }
}

resource "nomad_csi_volume_registration" "example" {
  plugin_id   = "hostpath-plugin0"
  volume_id   = "test-volume-reg"
  name        = "test-volume-reg"
  external_id = nomad_csi_volume.example.volume_id

  capability {
  }

  mount_options {
    fs_type = "ext4"
  }

  secrets_wo = jsonencode({
    "secret-key-1" = "secret-val-1"
  })

  topology_request {
    required {
      topology {
        segments = {
          "topology.hostpath.csi/node" = "node-0"
        }
      }
    }
  }
}

output "volume_secrets_wo_version" {
  value = nomad_csi_volume.example.secrets_wo_version
}

output "registration_secrets_wo_version" {
  value = nomad_csi_volume_registration.example.secrets_wo_version
}
```

#### Verification

1. `terraform apply` - both resources created, `secrets_wo_version` outputs `1`.
2. Change a value in `secrets_wo` and re-apply - `secrets_wo_version` increments to `2`.
3. Re-apply without changes - `secrets_wo_version` stays at `2` (no unnecessary update).
4. Confirm `secrets_wo` is not in `terraform.tfstate` (write-only).
5. `terraform destroy` - both resources removed cleanly.

[JIRA Ticket](https://hashicorp.atlassian.net/browse/NMD-1322)
Fixes #588


tgross:
Hi @vishnuedavarath! I'm going to review this. But first I pulled this down to do some testing and ran into a problem. I used the config you showed here (but without `nomad_csi_volume_registration.example`) because I didn't have that volume created yet. I created the volume and everything was fine. But when I update the `capacity_min` of the volume from 10GiB to 15GiB, the expand volume RPC is working on the plugin but then the provider is throwing an error.

Here's the plugin logs:

```
I0602 13:44:23.258058       1 server.go:101] GRPC call: /csi.v1.Controller/ControllerExpandVolume
I0602 13:44:23.258414       1 server.go:105] GRPC request: {"capacity_range":{"limit_bytes":21474836480,"required_bytes":16106127360},"secrets":"***stripped***","volume_capability":{"AccessType":{"Mount":{"fs_type":"ext4"}},"access_mode":{"mode":1}},"volume_id":"1a394447-5e89-11f1-bb40-c2be420b611c"}
I0602 13:44:23.258519       1 server.go:114] GRPC response: {"capacity_bytes":16106127360,"node_expansion_required":true}
I0602 13:44:23.258553       1 server.go:159] gRPCCall: {"Method":"/csi.v1.Controller/ControllerExpandVolume","Request":{"volume_id":"1a394447-5e89-11f1-bb40-c2be420b611c","capacity_range":{"required_bytes":16106127360,"limit_bytes":21474836480},"secrets":{"secret-key-1":"secret-val-1","secret-key-2":"secret-val-2"},"volume_capability":{"AccessType":{"Mount":{"fs_type":"ext4"}},"access_mode":{"mode":1}}},"Response":{"capacity_bytes":16106127360,"node_expansion_required":true},"Error":"","FullError":null}
```

Here's the result of the Terraform provider run:

```
$ terraform apply -auto-approve
╷
│ Warning: Provider development overrides are in effect
│
│ The following provider development overrides are set in the CLI configuration:
│  - hashicorp/nomad in /home/tim/go/bin
│
│ The behavior may therefore not match any released version of the provider and applying changes may
│ cause the state to become incompatible with published releases.
╵
nomad_csi_volume.example: Refreshing state... [id=test-volume]

Terraform used the selected providers to generate the following execution plan. Resource actions are
indicated with the following symbols:
  ~ update in-place

  # nomad_csi_volume.example will be updated in-place
  ~ resource "nomad_csi_volume" "example" {
      ~ capacity_min            = "10GiB" -> "15GiB"
        id                      = "test-volume"
        name                    = "test-volume"
        # (20 unchanged attributes hidden)

        # (3 unchanged blocks hidden)
    }

nomad_csi_volume.example: Modifying... [id=test-volume]
╷
│ Error: Provider produced inconsistent result after apply
│
│ When applying changes to nomad_csi_volume.example, provider
│ "provider[\"registry.terraform.io/hashicorp/nomad\"]" produced an unexpected new value: .capacity:
│ was cty.NumberIntVal(1.073741824e+10), but now cty.NumberIntVal(1.610612736e+10).
│
│ This is a bug in the provider, which should be reported in the provider's own issue tracker.
╵
╷
│ Error: Provider produced inconsistent result after apply
│
│ When applying changes to nomad_csi_volume.example, provider
│ "provider[\"registry.terraform.io/hashicorp/nomad\"]" produced an unexpected new value:
│ .capacity_min_bytes: was cty.NumberIntVal(1.073741824e+10), but now
│ cty.NumberIntVal(1.610612736e+10).
│
│ This is a bug in the provider, which should be reported in the provider's own issue tracker.
╵
```

website/docs/r/csi_volume_registration.html.markdown
```diff
@@ -14,8 +14,9 @@ This can be used to register and deregister CSI volumes in a Nomad cluster. The
 volume must already exist to be registered. Use the `nomad_csi_volume`
 resource to create a new volume.
 
-~> **Warning:** this resource will store any sensitive values placed in
-  `secrets` or `mount_options` in the Terraform's state file. Take care to
+~> **Warning:** The `secrets` attribute will store sensitive values in
+  Terraform's state file. Use `secrets_wo` instead to avoid storing secrets in
+  state. If you must use `secrets`, take care to
```
tgross: What about `mount_options`? (Same for the other resource)

internal/framework/provider/volumes/resource_csi_volume.go
```diff
@@ -0,0 +1,1274 @@
...
+		Description: "The storage API that will be used by the volume.",
+		Validators: []validator.String{
+			stringvalidator.OneOf(
+				"block-device",
+				"file-system",
+			),
+		},
+	}
+
+	if requiresReplace {
```
tgross: I realize this parameter is coming from the existing code, but I'm not sure if it's correct? I'm looking at the validation in [`CSIVolume.Create` RPC](https://github.com/hashicorp/nomad/blob/v2.0.2/nomad/csi_endpoint.go#L1112) and I _think_ we can change these values so long as the volume isn't currently claimed for a capability we've removed? Let's see if we can confirm that, because this would be a good time to fix the resource's behavior if so.

internal/framework/provider/volumes/resource_csi_volume.go
```diff
@@ -0,0 +1,1274 @@
...
+			Attributes: map[string]schema.Attribute{
+				"fs_type": schema.StringAttribute{
+					Optional:    true,
+					Description: "The file system type.",
+				},
+				"mount_flags": schema.ListAttribute{
+					ElementType: types.StringType,
+					Optional:    true,
+					Description: "The flags passed to mount.",
+				},
```
tgross: The API redacts mount flags because they can contain secrets (ex. for NFS mounts). While this isn't a great design of mount flags, it's an unfortunate reality of the `mount` syscalls. Should these be marked `sensitive`?

internal/framework/provider/volumes/resource_csi_volume_registration_test.go
```diff
@@ -0,0 +1,247 @@
+// Copyright IBM Corp. 2016, 2025
+// SPDX-License-Identifier: MPL-2.0
+
+package volumes_test
```
tgross, line 4: Needs an update test.

internal/framework/provider/volumes/resource_csi_volume.go
```diff
@@ -0,0 +1,1274 @@
...
+			ElementType:        types.StringType,
+			Optional:           true,
+			Sensitive:          true,
+			Description:        "An optional key-value map of strings used as credentials for publishing and unpublishing volumes. Deprecated: use secrets_wo instead.",
+			DeprecationMessage: "Use secrets_wo to avoid storing secrets in Terraform state.",
+		},
+		"secrets_wo": schema.StringAttribute{
+			Optional:    true,
+			WriteOnly:   true,
+			Description: "JSON-encoded map of secrets used as credentials for publishing and unpublishing volumes. Use jsonencode() to set this value. This value is write-only and will not be stored in Terraform state.",
```
tgross: This can't be a map that the provider encodes under the hood?

internal/framework/provider/volumes/resource_csi_volume.go
```diff
@@ -0,0 +1,1274 @@
...
+			return
+		}
+	}
+
+	// When the value is changing, pass config through as-is. State will store
+	// the user's exact string; semantic equality handles future no-op diffs.
+	resp.PlanValue = req.ConfigValue
+}
+
+func parseCapacity(capMinAttr, capMaxAttr types.String) (uint64, uint64, diag.Diagnostics) {
```
tgross: This is nitpicky, but we have a lot of functions in this resource file that are actually shared between the two CSI volumes resources (and probably could be shared with the DHV resource once that gets migrated). Should we pull these out into their own file in this directory?

website/docs/r/csi_volume_registration.html.markdown
vishnuedavarath: Since `mount_flags` are the sensitive part, would it make more sense to just add a `mount_flags_wo` write-only attribute inside the existing `mount_options` block instead of duplicating the entire block? `fs_type` values like "ext4" or "xfs" shouldn't contain secrets, so making those write-only seems unnecessary.

internal/framework/provider/volumes/resource_csi_volume.go
vishnuedavarath: Can `parameters` contain secrets too? Should we add a `parameters_wo` write-only alternative following the same pattern as `secrets_wo`, or is that out of scope since parameters aren't redacted by the Nomad API?

tgross: The CSI spec requires we treat parameters as opaque KV pairs ([ref](https://github.com/container-storage-interface/spec/blob/master/spec.md#createvolume)), and the same API that consumes them also has the secrets, which implies they're not going to be secret. I'd leave them be.

website/docs/r/csi_volume_registration.html.markdown
tgross: Ah, good catch. Agreed.

internal/framework/provider/volumes/resource_csi_volume_registration_test.go
vishnuedavarath: Added additional tests for in-place update and replace scenarios in csi-volume test.

tgross (dismissed):
LGTM

(Don't forget to request re-reviews when you're done replying to PR reviews!)

internal/framework/provider/volumes/resource_csi_volume.go
vishnuedavarath: Verified the behavior and  removed `requiresReplace` from capabilities. Updates in capabilities are now in-place, and `terraform apply` errors out when trying to remove a capability that's actively claimed.

tgross (approved):

PR #629, opened 2026-06-04
feat: add nomad_service and nomad_services data sources

### LINKED ISSUE #449: [Feature] Add a nomad_service data source.
Currently there is no data source to fetch services from nomad.  I'd like to be able to consume nomad services and push them into DNS, but I'd need a suitable data source to achieve this.

vishnuedavarath:
### Description
Implements data sources for Nomad's native service discovery API.

- nomad_service: reads registrations for a specific service name, with support for namespace, filter expressions, and rendezvous hashing (choose parameter)
- nomad_services: lists all registered services, with optional namespace filtering

Includes acceptance tests and website documentation.

#### Prerequisites

1. **Nomad cluster running** (dev mode is fine):
2. Terraform CLI (v1.0+)
3. Build the provider locally
4. Configure Terraform dev override 
5. Register a test service - run a job with `provider = "nomad"`

##### Test `nomad_services` data source
```terraform 
terraform {
  required_providers {
    nomad = {
      source = "hashicorp/nomad"
    }
  }
}

provider "nomad" {
}

data "nomad_services" "all" {}

data "nomad_services" "default_ns" {
  namespace = "default"
}
```

``` bash
terraform plan
```

##### Test `nomad_service` data source
```terraform
terraform {
  required_providers {
    nomad = {
      source = "hashicorp/nomad"
    }
  }
}

provider "nomad" {
  address = "http://127.0.0.1:4646"
}

# Basic lookup
data "nomad_service" "webapp" {
  service_name = "my-webapp"
}

# With filter
data "nomad_service" "canary_only" {
  service_name = "my-webapp"
  filter       = "\"canary\" in Tags"
}

# With choose (rendezvous hashing)
data "nomad_service" "one_instance" {
  service_name = "my-webapp"
  choose       = "1|my-key"
}
```

```bash
terraform plan
```

Fixes #449 
Fixes: https://hashicorp.atlassian.net/browse/NMD-1142

### DIFF
--- a/internal/framework/provider/services/datasource_service.go
+++ b/internal/framework/provider/services/datasource_service.go
@@ -204,10 +204,11 @@ func (d *ServiceDataSource) Read(ctx context.Context, req datasource.ReadRequest
 		return
 	}
 
-	data.ID = types.StringValue(serviceName)
 	if data.Namespace.IsNull() || data.Namespace.IsUnknown() {
 		data.Namespace = types.StringValue("default")
 	}
+
+	data.ID = types.StringValue(data.Namespace.ValueString() + "/" + serviceName)
 
 	regList := make([]attr.Value, 0, len(registrations))
 	for _, reg := range registrations {
--- a/internal/framework/provider/services/datasource_service.go
+++ b/internal/framework/provider/services/datasource_service.go
@@ -1,0 +1,255 @@
+// Copyright IBM Corp. 2016, 2026
+// SPDX-License-Identifier: MPL-2.0
+
+package services
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/hashicorp/nomad/api"
+	"github.com/hashicorp/terraform-plugin-framework/attr"
+	"github.com/hashicorp/terraform-plugin-framework/datasource"
+	"github.com/hashicorp/terraform-plugin-framework/datasource/schema"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+	"github.com/hashicorp/terraform-plugin-log/tflog"
+	"github.com/hashicorp/terraform-provider-nomad/nomad"
+)
+
+var _ datasource.DataSource = &ServiceDataSource{}
+var _ datasource.DataSourceWithConfigure = &ServiceDataSource{}
+
+type ServiceDataSource struct {
+	providerConfig nomad.ProviderConfig
+}
+
+func NewServiceDataSource() datasource.DataSource {
+	return &ServiceDataSource{}
+}
+
+type serviceModel struct {
+	ID            types.String `tfsdk:"id"`
+	ServiceName   types.String `tfsdk:"service_name"`
+	Namespace     types.String `tfsdk:"namespace"`
+	Filter        types.String `tfsdk:"filter"`
+	Choose        types.String `tfsdk:"choose"`
+	Registrations types.List   `tfsdk:"registrations"`
+}
+
+var registrationObjectType = types.ObjectType{
+	AttrTypes: map[string]attr.Type{
+		"id":           types.StringType,
+		"address":      types.StringType,
+		"port":         types.Int64Type,
+		"node_id":      types.StringType,
+		"datacenter":   types.StringType,
+		"alloc_id":     types.StringType,
+		"job_id":       types.StringType,
+		"namespace":    types.StringType,
+		"tags":         types.ListType{ElemType: types.StringType},
+		"create_index": types.Int64Type,
+		"modify_index": types.Int64Type,
+	},
+}
+
+func (d *ServiceDataSource) Metadata(_ context.Context, req datasource.MetadataRequest, resp *datasource.MetadataResponse) {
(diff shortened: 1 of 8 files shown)


website/docs/d/services.html.markdown
```diff
@@ -0,0 +1,55 @@
...
+  namespace = "*"
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+- `namespace` `(string: <optional>)` - The namespace to filter services. Use
+  `"*"` to list services across all namespaces.
```
tgross: Should `"*"` be the default, rather than empty, which IIRC gets turned into `"default"` by the API? If we're not going to default to `"*"`, we should at least document what the default behavior is.

website/docs/d/services.html.markdown
```diff
@@ -0,0 +1,55 @@
...
+layout: "nomad"
+page_title: "Nomad: nomad_services"
+sidebar_current: "docs-nomad-datasource-services"
+description: |-
+  Retrieve the list of all registered Nomad services.
+---
+
+# nomad_services
+
+Retrieve the list of all registered Nomad services.
```
tgross, line 11: I was going to ask why we don't have filter arguments here but apparently we don't support them on the API! Seems like an oversight to be fixed in core.

website/docs/d/service.html.markdown
```diff
@@ -0,0 +1,75 @@
...
+  [expression][nomad_api_filter] used to filter the results.
+- `choose` `(string: <optional>)` - Specifies the number of services to return
+  and a hash key for rendezvous hashing. Must be in the form
+  `"<number>|<key>"`.
+
+## Attribute Reference
+
+The following attributes are exported:
+
+- `id` `(string)` - The service name used as the data source ID.
```
tgross: This description isn't distinct enough from the `registrations[].id` for the end user, I suspect. Maybe call out that this is how you refer to this entire data source.

(This comment and the one below apply to the help text in the code as well)

website/docs/d/service.html.markdown
```diff
@@ -0,0 +1,75 @@
...
+- `registrations` `(list of registrations)` - A list of service registrations
+  matching the query. Each registration has the following attributes:
+  - `id` `(string)` - The unique identifier of the service registration.
```
tgross: Further distinction from the above:

```suggestion
  - `id` `(string)` - The unique identifier of the service registration for a specific allocation.
```

internal/framework/provider/services/datasource_services_test.go
```diff
@@ -0,0 +1,146 @@
...
+
+	// Wait for the service to be registered.
+	deadline := time.Now().Add(30 * time.Second)
+	for time.Now().Before(deadline) {
+		services, _, err := client.Services().Get(serviceName, &api.QueryOptions{Namespace: namespace})
+		if err == nil && len(services) > 0 {
+			break
+		}
+		time.Sleep(500 * time.Millisecond)
+	}
```
tgross, line 170: Where's the assertion?

internal/framework/provider/services/datasource_services.go
```diff
@@ -0,0 +1,178 @@
...
+	serviceStubs, _, err := client.Services().List(qOpts)
+	if err != nil {
+		resp.Diagnostics.AddError(
+			"Error listing Nomad services",
+			fmt.Sprintf("Unable to list services: %s", err),
+		)
+		return
+	}
+
+	data.ID = types.StringValue("nomad-services")
```
tgross: Doesn't this ID need to be namespaced? What happens if I have multiple data sources, each with a different namespace config?

internal/framework/provider/services/datasource_service.go
```diff
@@ -0,0 +1,255 @@
...
+	registrations, _, err := client.Services().Get(serviceName, qOpts)
+	if err != nil {
+		resp.Diagnostics.AddError(
+			"Error reading Nomad service",
+			fmt.Sprintf("Unable to read service %q: %s", serviceName, err),
+		)
+		return
+	}
+
+	data.ID = types.StringValue(serviceName)
```
tgross: This ID needs to be namespaced, because the service is namespaced and can be duplicated in multiple namespaces.

tgross (changes requested):

internal/framework/provider/services/datasource_services.go
vishnuedavarath: I'm removing the id attribute as this is a new data source in the Plugin Framework provider, and Framework-based resources/data sources have no mandate to store state in id unlike SDKv2 resources/data sources. Since we don't have any real ID to store here, I'm removing it.

website/docs/d/services.html.markdown
vishnuedavarath: Keeping it as "default" to match the Nomad API behavior, the API defaults to the "default" namespace when none is provided. Updated the docs to make this explicit so users would know they need `namespace = "*"` for cross-namespace listing.

tgross (approved):
LGTM

PR #630, opened 2026-06-08
Fix dynamic_host_volume persisting capacity changes to state after failure

### LINKED ISSUE #619: dynamic_host_volume persists changes to capacity after failure
### Affected Resource(s)

- dynamic_host_volume
- data_source_dynamic_host_volume

```text
07:00:04.013 STDOUT terraform:   ~ resource "nomad_dynamic_host_volume" "default" {
07:00:04.013 STDOUT terraform:       ~ capacity_max       = "2GiB" -> "3GiB"
07:00:04.014 STDOUT terraform:       ~ capacity_min       = "2GiB" -> "3GiB"
07:00:04.014 STDOUT terraform:         id                 = "74a165e1-17cc-1f1a-2948-21cb3b08408f"
07:00:04.014 STDOUT terraform:         name               = "my-volume"
07:00:04.014 STDOUT terraform:         # (11 unchanged attributes hidden)
07:00:04.014 STDOUT terraform:         # (2 unchanged blocks hidden)
07:00:04.014 STDOUT terraform:     }
```

I'm using a custom plugin on the Nomad side which forbids changing the size of an existing volume as it is something I currently don't support. The script correctly exits 1 in that case: 

```bash
if [ "${curr_size}" -ne "${size}" ]; then
  stderr "size mismatch: loop ${loop_dev} is ${curr_size} bytes != requested ${size} bytes"
  stderr "growing or shrinking the volume is not supported"
  exit 1
fi
```

### Expected Behavior

On Terraform apply the run should fail with error and the same should happen on subsequent applies.

### Actual Behavior

The first apply fails but the Terraform state is updated with the requested capacity and subsequent runs complete without errors.

Terraform state after the first run:

```
terraform state show 'nomad_dynamic_host_volume.default["my-volume"]'
# nomad_dynamic_host_volume.default["my-volumer"]:
resource "nomad_dynamic_host_volume" "default" {
    capacity           = "2.0 GiB"
    capacity_bytes     = 2147483648
    capacity_max       = "3GiB"      <---- this should be 2GiB
    capacity_max_bytes = 2147483648
    capacity_min       = "3GiB"      <---- this should be 2GiB
    capacity_min_bytes = 2147483648
```

The issue originates from here: https://github.com/hashicorp/terraform-provider-nomad/blame/main/nomad/data_source_dynamic_host_volume.go#L178

vishnuedavarath:
When updating a dynamic host volume with SDKv2, the planned config values (e.g. capacity_min, capacity_max) leak into Terraform state even when the API call fails. On subsequent refreshes, the Read function was conditionally skipping these fields, causing state to remain inconsistent.

Changes:
- Always set capacity_min and capacity_max from the API response in Read, ensuring state reflects the actual remote resource
- Add DiffSuppressFunc for capacity_min and capacity_max to suppress spurious diffs from equivalent byte representations (e.g. "1GiB" vs "1.0 GiB")
- Refactor dynamicHostVolumeRead into dynamicHostVolumeReadImpl with a flag to conditionally set capacity string fields, avoiding panics in the registration resource which lacks these schema fields

**Setup:** A custom Nomad host volume plugin that rejects capacity changes:

```bash
#!/bin/bash
# no-resize plugin - rejects any size change on existing volumes
if [ "${curr_size}" -ne "${size}" ]; then
  echo "growing or shrinking the volume is not supported" >&2
  exit 1
fi
```

#### Reproduction (before fix):
1. Create a dynamic host volume:
``` terraform
resource "nomad_dynamic_host_volume" "test" {
  name         = "my-volume"
  plugin_id    = "no-resize"
  capacity_min = "1GiB"
  capacity_max = "1GiB"

  capability {
  }
}
```
2. Run `terraform apply` - succeeds, volume created at 1GiB.
3. Update `capacity_min = "2GiB"` and `capacity_max = "2GiB"`.
4. Run `terraform apply` - fails (plugin exits 1, rejects resize).
5. Run `terraform apply` again with the same 2GiB values - succeeds with no error and no changes. The state has the requested 2GiB values even though the volume is still 1GiB. Terraform believes the resource is in sync and never retries the update.

#### Expected behavior (after fix):
- Step 4: Apply fails. On next `terraform plan`/`terraform apply`, the refresh detects drift - state had stale "2GiB" from the failed apply, but the API reports "1.0 GiB". Terraform corrects the state.
- Step 5: Subsequent applies with 2GiB in config correctly show a diff (state "1.0 GiB" → config "2GiB") and attempt the update again (which will fail again from the plugin).

#### Verifying diff suppression for equivalent formats

1. Create a volume with capacity_min = "1GiB".
2. Change config to capacity_min = "1.0 GiB" or capacity_min = "1024MiB".
3. Run `terraform plan` - should show no changes (same byte value).

Fixes: #619
Fixes: https://hashicorp.atlassian.net/browse/NMD-1425

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ IMPROVEMENTS:
 BUG FIXES:
 * data source/nomad_variable: Fix panic when reading a variable due to `items_wo_version` not being in the data source schema. ([#625](https://github.com/hashicorp/terraform-provider-nomad/pull/625))
 * resource/nomad_acl_token: Fixed perpetual destroy-and-recreate cycle when `expiration_ttl` is set to a duration like `"1h"` or `"30m"`. ([#615](https://github.com/hashicorp/terraform-provider-nomad/pull/615))
+* resource/nomad_dynamic_host_volume: Fix `capacity_min` and `capacity_max` being incorrectly persisted to state after a failed update, and suppress spurious diffs from equivalent capacity representations (e.g. `"1GiB"` vs `"1.0 GiB"`). ([#630](https://github.com/hashicorp/terraform-provider-nomad/pull/630))
 
 ## 2.6.1 (April 20, 2026)
 
--- a/nomad/data_source_dynamic_host_volume.go
+++ b/nomad/data_source_dynamic_host_volume.go
@@ -14,7 +14,7 @@ import (
 
 func dataSourceDynamicHostVolume() *schema.Resource {
 	return &schema.Resource{
-		Read: dynamicHostVolumeRead,
+		Read: dynamicHostVolumeReadWithCapacity,
 		Schema: map[string]*schema.Schema{
 			"id": {
 				Description: "Volume ID",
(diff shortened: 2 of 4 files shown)


tgross (approved):
LGTM!

PR #631, opened 2026-06-17
job: get submission from core parser instead of building one

tgross:
In https://github.com/hashicorp/nomad/pull/28138 we're expanding the HCL2 parser to have a method that can return an `api.JobSubmission` object with the correctly-parsed variables. This is to prevent `-var` flags from being encoded as strings if they actually represent complex values, which then breaks in the web UI editor.

Update the job resource to use this same parser. And expand the job resource fields to add `var_content` so that users can use blobs of HCL-formatted variables just as they do with the command line, which may be better-suited for complex types.

Ref: https://github.com/hashicorp/nomad/pull/28138
Ref: https://hashicorp.atlassian.net/browse/NMD-531


tgross:
I've been working on the panic here and don't yet have an answer as to why this is broken. WIP.

PR #632, opened 2026-07-02
resource/nomad_job: add preserve_resources argument

vishnuedavarath:
Support the Nomad API's PreserveResources register option via a new preserve_resources argument. Includes docs update and acceptance test.

### Description

Adds a `preserve_resources` argument to the `nomad_job` resource. When set to true, the provider passes the Nomad API's `PreserveResources` register option, so that task resources (cpu/memory) already running in the cluster are preserved during a job update instead of being reset to the values in the jobspec.

1. `terraform init`
2. Register a job
```hcl
job "demo" {
  datacenters = ["dc1"]
  group "api" {
    task "server" {
      driver = "raw_exec"
      config { command = "/bin/sleep"; args = ["600"] }
      resources { cpu = 100; memory = 32 }
    }
  }
}
```

```hcl
resource "nomad_job" "demo" {
  jobspec            = file("job.nomad")
}
```
3. Simulate an out-of-band change to job's task-group resources
4. `terraform apply` with `preserve_resources` set
```hcl
resource "nomad_job" "demo" {
  preserve_resources = true
  jobspec            = file("job.nomad")
}
```

JIRA Ticket: https://hashicorp.atlassian.net/browse/NMD-1575


jrasell (approved):
LGTM!

Issue #647, opened 2026-09-16
job modify index has changed since last refresh — triggered by nomad autoscaler


econsult-devops:
We are experiencing an issue similar to #398 . It's caused by the Nomad Autoscaler and our pipeline. Here is the gist of it:
a plan is generated in one stage and applied in a separate, later stage. 
If Nomad Autoscaler scales the job's task_group during that window, the job's modify index changes before the saved plan is applied, causing the apply step to fail with `job modify index has changed since last refresh`.

### Terraform Version

```
Terraform v1.11.1
on linux_amd64
+ provider registry.terraform.io/hashicorp/nomad v2.6.1
+ provider registry.terraform.io/mecodia/pass v3.1.0
```

### Nomad Version

3 node server cluster at version 1.11.3

### Provider Configuration

```hcl
provider "nomad" {
  address   = ""
  secret_id = ""
}
```

### Environment Variables

N/A

### Affected Resource(s)

- nomad_job

### Terraform Configuration Files

```hcl
resource "nomad_job" "this" {
  jobspec = file("job.nomad.hcl")

  detach           = false
  purge_on_destroy = true
  preserve_counts  = true
}
```

```hcl
job "whoami" {
  node_pool = "test"

  spread {
    attribute = "$${node.datacenter}"
  }
  vault {}

  group "whoami" {
    volume "certs" {
      type      = "host"
      source    = "certs"
      read_only = true
    }

    scaling {
      enabled = true
      min     = 1
      max     = 10
    }

    network {
      mode = "bridge"
      port "http" {}
      dns {
        servers = ["${attr.unique.network.ip-address}"]
      }
    }

    service {
      name = "whoami"
      port = "http"
      tags = [
        "http",
        "traefik.enable=true",
        "traefik.http.routers.whoami.entrypoints=external",
        "node_pool=$${meta.node_pool}",
      ]
    }

    task "whoami" {
      driver = "podman"
      config {
        image = "private.registry.io/docker/whoami:v1.12"
        ports = ["http"]
      }

      env {
        WHOAMI_PORT_NUMBER = "${NOMAD_PORT_http}"
        WHOAMI_NAME        = "amiwho"
      }

      resources {
        memory     = 64
        memory_max = 64
      }
    }
  }
}
```

### Debug Output

N/A

### Panic Output

N/A

### Expected Behavior

[...]

PR #648, opened 2026-09-16
add skip_modify_index_check to nomad_job resource to resolve issue #647

econsult-devops:
### Description
Hi, as the title mentions this a pr to resolve the issue described in #647 
How to test and reproduce the bug is also mentioned there.

Please let us know if you require any updates or tests before considering merging the PR


nomad/resource_job.go
```diff
@@ -70,6 +70,13 @@ func resourceJob() *schema.Resource {
 				Type:        schema.TypeBool,
 			},
 
+			"skip_modify_index_check": {
+				Description: "If true does not check modify index during job updates. This disables drift detection in nomad itself. Only use this if you only apply job changes via terraform.",
```
tgross, line 74: > Only use this if you only apply job changes via terraform.

I'm confused... isn't the underlying problem that you _are_ changing the job outside of Terraform? This isn't safe to do, is it?

tgross (changes requested):

nomad/resource_job.go
econsult-devops: The description is kinda lacking, that's on me. 
Job changes only happen through two paths: Terraform, or the Nomad autoscaler. The autoscaler changes are strictly limited to scaling actions (e.g. count) and nothing else, which is why this flag is safe to use in that scenario. I can update the description to make that clearer.

